PR #6, opened 2021-01-08
Ensure the host can be parsed as an IPv6 address.

gabi-250:
If the host is an IPv6 address, the value returned by `Uri::host()` will [include the surrounding brackets](https://github.com/hyperium/http/pull/292). This causes boring to fail to parse the host as an [IP address](https://github.com/cloudflare/boring/blob/5cb8947d7ec6e0c67fa40167940f313b8d6537f8/boring/src/ssl/connector.rs#L342). Eventually, this causes hostname verification to fail. I _assume_ the hostname verification failure is due to the fact that the host will be compared with the `DNSName`s from `SubjectAlternativeName`, instead of the `IPAddress`es ([`set_ip` vs `set_host`](https://github.com/cloudflare/boring/blob/5cb8947d7ec6e0c67fa40167940f313b8d6537f8/boring/src/ssl/connector.rs#L341-L342)).

### DIFF
--- a/hyper-boring/src/lib.rs
+++ b/hyper-boring/src/lib.rs
@@ -21,6 +21,7 @@ use std::fmt::Debug;
 use std::future::Future;
 use std::io;
 use std::mem::MaybeUninit;
+use std::net;
 use std::pin::Pin;
 use std::sync::Arc;
 use std::task::{Context, Poll};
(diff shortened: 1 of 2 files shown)


cbranch (approved):

hyper-boring/src/lib.rs
```diff
@@ -230,8 +230,17 @@ where
 
             let host = uri.host().ok_or("URI missing host")?;
 
-            let config = inner.setup_ssl(&uri, host)?;
-            let stream = tokio_boring::connect(config, host, conn).await?;
+            // If `host` is an IPv6 address, we must strip away the square brackets that surround
+            // it (otherwise, boring will fail to parse the host as an IP address, eventually
+            // causing the handshake to fail due a hostname verification error).
+            let host_or_ip = match (host.find('['), host.find(']')) {
+                (Some(0), Some(i)) => &host[1..i],
```
inikulin, line 237: We need to ensure here that `i == host.len() - 1`. Otherwise, we can strip a part of host, doesn't seem to be a safe thing to do. Should be smthg like:
```rust
let last = host.len() - 1;

if last > 0 && host[0] == '[' && host[last] == ']' {
    host = host [1..last];
}
```
Ideally we should also validate that what's enclosed in `[ ]` is indeed and IPv6

inikulin (changes requested):

hyper-boring/src/lib.rs
gabi-250: Good point (`host` is extracted from a valid `Uri`, so I assumed it could only contain brackets if it is a valid IPv6 address). That being said I'm happy to make this change.

gabi-250:
@inikulin Thank you for the review. I made the changes you've suggested.

inikulin (approved):

PR #7, opened 2021-01-28
Include .asm files required for Windows

hayleyjames:
When this crate is used as a dependency on Windows, boring-sys fails to build because Cargo doesn't include .asm files.

This problem was previously seen in [quiche](https://github.com/cloudflare/quiche/pull/776).


inikulin (approved):
Thanks!

inikulin:
Published to crates.io

Issue #10, opened 2021-02-10
bindgen usage for boring-sys


jedisct1:
Hi,

And thanks for these crates!

The BoringSSL revision is constantly kept up to date, which is really neat. However, declarations from `boring-sys` are running behind the BoringSSL API changes.

Could it be an option to use `bindgen` for `boring-sys` in order to automatically keep the external declarations in sync with the BoringSSL API?

This is something I can contribute, but maybe there is a reason why bindings seem to be manually maintained rather than leveraging `bindgen`.

Thanks!

inikulin:
Hi, thank you!

The project has started as a fork of rust-openssl, which supports few different versions of OpenSSL and LibreSSL, so it was impossible to use bindgen there. However, as you pointed, we aim to provide API only for latest BoringSSL version and, hence, bindgen-based approach will work for us. PRs are welcome 😃

nox:
I think this can be closed.

inikulin:
Implemented in https://github.com/cloudflare/boring/pull/11

PR #11, opened 2021-02-11
Use bindgen to automatically generate boring-sys

jedisct1:
This ensures that all the Rust functions, types and constants always match the actual BoringSSL definitions.

It also removes quite a lot of manually maintained code, as well as the need for `systest`.

The value for `SslOptions::ALL`, for example, was wrong. On current BoringSSL versions, this is a no-op, and is set to `0`.

Clearing it does nothing. So, the `clear_ctx_options` test, that passed by accident, was adjusted to use a different option.

The `libc` crate is not required, as we only use it for types that are already defined in the standard library. It was removed from `boring-sys`. The same can be done to other crates later.


jedisct1:
Fixes #10

inikulin:
This is awesome, thanks! 

Can we keep `systest` to ensure generated bindings validity? Also, it seems like it fails to compile on Windows

boring-sys/src/lib.rs
```diff
@@ -95,7 +76,7 @@ pub fn init() {
 
     let init_options = OPENSSL_INIT_LOAD_SSL_STRINGS;
 
-    INIT.call_once(|| unsafe {
-        OPENSSL_init_ssl(init_options, ptr::null_mut());
-    })
+    INIT.call_once(|| {
+        assert!(unsafe { OPENSSL_init_ssl(init_options.try_into().unwrap(), ptr::null_mut()) } == 1)
```
inikulin, line 80: Nit:  `assert_eq!` will be less error prone

boring/src/ssl/mod.rs
```diff
@@ -110,71 +110,71 @@ bitflags! {
     /// Options controlling the behavior of an `SslContext`.
     pub struct SslOptions: c_uint {
         /// Disables a countermeasure against an SSLv3/TLSv1.0 vulnerability affecting CBC ciphers.
-        const DONT_INSERT_EMPTY_FRAGMENTS = ffi::SSL_OP_DONT_INSERT_EMPTY_FRAGMENTS;
+        const DONT_INSERT_EMPTY_FRAGMENTS = ffi::SSL_OP_DONT_INSERT_EMPTY_FRAGMENTS as _;
 
         /// A "reasonable default" set of options which enables compatibility flags.
-        const ALL = ffi::SSL_OP_ALL;
+        const ALL = ffi::SSL_OP_ALL as _;
```
inikulin, line 116: What type it currently generates for `#define`'s?

boring-sys/build.rs
```diff
@@ -191,4 +194,61 @@ fn main() {
...
+        "ossl_typ.h",
+        "pkcs12.h",
+        "poly1305.h",
+        "rand.h",
+        "rc4.h",
+        "ripemd.h",
+        "siphash.h",
+        "srtp.h",
+        "trust_token.h",
+        "x509v3.h",
```
inikulin, line 243: Do we want to maintain this list manually or we can just glob by `*.h` instead now?

boring-sys/build.rs
```diff
@@ -191,4 +194,61 @@ fn main() {
...
+    let mut builder = bindgen::Builder::default()
+        .derive_copy(true)
+        .derive_debug(true)
+        .derive_default(true)
+        .derive_eq(true)
+        .default_enum_style(bindgen::EnumVariation::NewType { is_bitfield: false })
+        .default_macro_constant_type(bindgen::MacroTypeVariation::Signed)
+        .generate_comments(true)
+        .fit_macro_constants(false)
+        .size_t_is_usize(true)
```
inikulin, line 208: Even though it's true for the most modern systems, not sure it's always a safe assumption.

boring-sys/src/lib.rs
jedisct1: Sure!

boring/src/ssl/mod.rs
jedisct1: The default type is `i32` (unless a constant requires `i64`).

That can be changed to `u32` or even to the smallest type that can represent a given value, but `i32` is what requires the least amount of conversions.

boring-sys/build.rs
jedisct1: Globbing `*.h` would fail due to architecture-specific headers (`arm_arch.h`).

This list is the minimum set of files to covert the entire API. We can glob and blacklist some files, but as some of these headers are not meant to be directly included by applications, that may not be reliable either.

boring-sys/build.rs
jedisct1: This is true on all platforms Rust currently works on and Rust doesn't have a `size_t` type.

That recent change in `bindgen` is still being [subject to discussion](https://github.com/rust-lang/rust-bindgen/issues/1901). The previous behavior remains the most intuitive and actually safest way for bindings to interact with C types.

jedisct1:
The Windows CI issue is due to https://github.com/KyleMayes/clang-sys/issues/121

Should be good now!

inikulin (approved):

jedisct1:
> Can we keep systest to ensure generated bindings validity?

`bindgen` is configured to generate layout tests for every single type it defines, so `systest` seems a bit redundant. `systest` also fails to parse the generated bindings due to a current bug in the `syntex` crate it depends on.

inikulin:
@cbranch @ghedo @wbl could you take a look too pls?

wbl (approved):

PR #13, opened 2021-02-19
Bind the select_cert hook and allow users to inspect extensions early

nox:


boring/src/ssl/test/mod.rs
```diff
@@ -480,6 +480,76 @@ fn test_alpn_server_unilateral() {
...
+        }
+    });
+    let server = server.build();
+
+    let mut client = server.client();
+    client.ctx().set_alpn_protos(b"\x06http/2").unwrap();
+    client.connect();
+    assert_eq!(
+        alpn_extension.lock().unwrap().as_deref(),
+        Some(&b"\x00\x07\x06http/2"[..])
```
nox, line 549: Note that it returns the entire extension data, including the leading length, not sure that's good or not.

boring/src/ssl/mod.rs
```diff
@@ -1560,6 +1588,24 @@ pub struct CipherBits {
...
+#[repr(transparent)]
+pub struct ClientHello(ffi::SSL_CLIENT_HELLO);
+
+impl ClientHello {
+    pub fn get_extension(&self, type_: u16) -> Option<&[u8]> {
+        unsafe {
+            let mut ptr = ptr::null();
+            let mut len = 0;
+            let result =
+                ffi::SSL_early_callback_ctx_extension_get(&self.0, type_, &mut ptr, &mut len);
```
inikulin, line 1600: Not sure that we're dealing correctly with the memory here: we let Boring init the ptr, but it later will be dropped by Rust

boring/src/ssl/test/mod.rs
```diff
@@ -480,6 +480,76 @@ fn test_alpn_server_unilateral() {
...
+#[test]
+fn test_select_cert_alpn_extension() {
+    let mut server = Server::builder();
+    let alpn_extension = std::sync::Arc::new(std::sync::Mutex::new(None));
+    server.ctx().set_select_certificate_callback({
+        let alpn = alpn_extension.clone();
+        move |client_hello| {
+            *alpn.lock().unwrap() = Some(
+                client_hello
+                    .get_extension(ffi::TLSEXT_TYPE_application_layer_protocol_negotiation as u16)
```
inikulin, line 535: Let's expose extension type enum and use it instead of raw ints

boring/src/ssl/test/mod.rs
inikulin: Yeah, better strip the leading length

boring/src/ssl/mod.rs
nox: I am not sure what you mean:

> Otherwise it sets out_data to point to the extension contents (not including the type and length bytes), sets out_len to the length of the extension contents and returns one.

We borrow `self` to return the slice, so it's not like we can mutate the `ClientHello` value and bork the slice contents.

boring/src/ssl/test/mod.rs
nox: I don't think there is an extension type enum, the cast is here because the FFI constant is i32.

boring/src/ssl/test/mod.rs
nox: I've stripped the length and made the function panic if that doesn't correspond to the length of the rest of the slice. AFAIK the handshake wouldn't have reached the point where you can call that function if the lengths didn't correspond.

boring/src/ssl/test/mod.rs
nox: I added a custom fake-enum struct type for the extension type.

inikulin: Yeah, this is what I've meant - it's done in a few other places where integer variant enum in introduced in Rust API for the convenience

boring/src/ssl/mod.rs
```diff
@@ -463,6 +463,56 @@ impl AlpnError {
...
+/// An error returned from a certificate selection callback.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct SelectCertError(ffi::ssl_select_cert_result_t);
+
+impl SelectCertError {
+    /// A fatal error occured and the handshake should be terminated.
+    pub const ERROR: Self = Self(ffi::ssl_select_cert_result_t::ssl_select_cert_error);
+}
+
+/// An error returned from a certificate selection callback.
```
inikulin, line 475: Need a different comment here

boring/src/ssl/mod.rs
```diff
@@ -463,6 +463,56 @@ impl AlpnError {
...
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct ExtensionType(u16);
+
+impl ExtensionType {
```
inikulin, line 479: Let's expose `From<u16>` for custom extensions

boring/src/ssl/mod.rs
```diff
@@ -1560,6 +1629,33 @@ pub struct CipherBits {
...
+#[repr(transparent)]
+pub struct ClientHello(ffi::SSL_CLIENT_HELLO);
+
+impl ClientHello {
+    /// Returns the data of a given extension, if present.
+    ///
+    /// This corresponds to [`SSL_early_callback_ctx_extension_get`].
+    ///
+    /// [`SSL_early_callback_ctx_extension_get`]: https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_early_callback_ctx_extension_get
+    pub fn get_extension(&self, type_: ExtensionType) -> Option<&[u8]> {
```
inikulin, line 1641: Let's rename the var to `ext_type`. Not a big fan of the `_` workaround tbh

boring/src/ssl/mod.rs
```diff
@@ -1560,6 +1629,33 @@ pub struct CipherBits {
...
+        let slice = unsafe {
+            let mut ptr = ptr::null();
+            let mut len = 0;
+            let result =
+                ffi::SSL_early_callback_ctx_extension_get(&self.0, type_.0, &mut ptr, &mut len);
```
inikulin, line 1646: Ah, it's `const uint8_t **`. I thought it's `uint8_t *` which we need to de-alloc ourselves.

boring/src/ssl/mod.rs
```diff
@@ -1560,6 +1629,33 @@ pub struct CipherBits {
...
+            if result == 0 {
+                return None;
+            }
+            slice::from_raw_parts(ptr, len)
+        };
+        let (len, rest) = slice.split_at(2);
+        let len = (len[0] as usize) << 8 | len[1] as usize;
```
inikulin, line 1653: `u16::from_be_bytes([len[0], len[1]])`?

boring/src/ssl/mod.rs
```diff
@@ -1560,6 +1629,33 @@ pub struct CipherBits {
...
+        assert_eq!(len, rest.len());
```
inikulin, line 1654: Not sure if we want to panic here, considering we're working with the user-provided data. We should check for the length prefix to be exactly 2 bytes and that this invariant holds and return `Err(MalformedExtension)` or smthg, if it's not

boring/src/ssl/test/mod.rs
```diff
@@ -480,6 +481,78 @@ fn test_alpn_server_unilateral() {
...
+
+    let client = server.client();
+    client.connect_err();
+}
+
+#[test]
+fn test_select_cert_unknown_extension() {
+    let mut server = Server::builder();
+    let unknown_extension = std::sync::Arc::new(std::sync::Mutex::new(None));
+    server.ctx().set_select_certificate_callback({
```
inikulin, line 513: Style nit: here and in all other places: separate var declarations from multiline expressions/statements by a blank line

inikulin (changes requested):

boring/src/ssl/mod.rs
inikulin: Ah, it's const uint8_t **. I thought it's uint8_t * which we need to de-alloc ourselves.

boring/src/ssl/mod.rs
nox: Thinking twice about it I feel like this is dependent on the expected type of the extension data of each extension and it should probably not be trimmed here, I'll check the various specs to confirm.

nox: First, [RFC 5246](https://tools.ietf.org/html/rfc5246#section-4.3) says:

> A vector (single-dimensioned array) is a stream of homogeneous data elements.  The size of the vector may be specified at documentation time or left unspecified until runtime. In either case, the length declares the number of bytes, not the number of elements, in the vector.

Then it says:

> Variable-length vectors are defined by specifying a subrange of legal lengths, inclusively, using the notation `<floor..ceiling>`.  When these are encoded, the actual length precedes the vector's contents in the byte stream.  The length will be in the form of a number consuming as many bytes as required to hold the vector's specified maximum (`ceiling`) length. A variable-length vector with an actual length field of zero is referred to as an empty vector.

So I didn't know that the lengths of arrays in those RFCs are in terms of *bytes* not in terms of the number of elements. This is why I assumed that the length we are now stripping was the length of the opaque payload of the extension. I can now see that this is wrong.

Furthermore, [RFC 7301](https://tools.ietf.org/html/rfc7301#section-3.1) says:

>    The "extension_data" field of the ("application_layer_protocol_negotiation(16)") extension SHALL contain a "ProtocolNameList" value.
> 
> ```
> opaque ProtocolName<1..2^8-1>;
> 
> struct {
>     ProtocolName protocol_name_list<2..2^16-1>
> } ProtocolNameList;
> ```

So that length is indeed always 2 bytes in the case of the ALPN extension, and we shouldn't be stripping it as this is not an ALPN-specific method.

nox: I have reverted the trimming as this is not the place where we should be doing that.

inikulin (approved):

PR #16, opened 2021-03-01
Make Display for Error more readable

nox:


gabi-250 (approved):

boring/src/error.rs
```diff
@@ -251,25 +251,11 @@ impl fmt::Debug for Error {
...
-        }
         write!(
             fmt,
-            ":{}:{}:{}",
-            self.file(),
-            self.line(),
-            self.data().unwrap_or("")
+            "{} ({})",
+            self.reason().unwrap_or("unknown error"),
+            self.code()
```
inikulin, line 258: Would like to keep file/line info. It was very useful previously.

nox: They are already printed in the `Debug` impl though.

inikulin (approved):

PR #17, opened 2021-03-04
Improve error printing

nox:


boring/src/ssl/error.rs
```diff
@@ -152,21 +153,42 @@ impl<S: fmt::Debug> StdError for HandshakeError<S> {
...
-            HandshakeError::SetupFailure(ref e) => write!(f, "stream setup failed: {}", e)?,
-            HandshakeError::Failure(ref s) => {
-                write!(f, "the handshake failed: {}", s.error())?;
-                let verify = s.ssl().verify_result();
-                if verify != X509VerifyResult::OK {
-                    write!(f, ": {}", verify)?;
-                }
+        let (prefix, s) = match *self {
+            HandshakeError::SetupFailure(ref e) => {
+                return write!(f, "TLS stream setup failed:\n\n{}", e)
```
inikulin, line 158: This is kind of indirection which can cause errors in the future. Let's get rid of this early return

boring/src/ssl/error.rs
```diff
@@ -152,21 +153,42 @@ impl<S: fmt::Debug> StdError for HandshakeError<S> {
...
+            }
+            HandshakeError::Failure(ref s) => ("TLS handshake failed", s),
+            HandshakeError::WouldBlock(ref s) => ("TLS handshake interrupted", s),
+        };
+        let verify = s.ssl().verify_result();
+        if verify != X509VerifyResult::OK {
```
inikulin, line 164: Nit: try to avoid negations in `if` branch condition. Inverse it, so the `else` contains the code for the cases when `verify != OK`. Also, verify variable is excessive here

boring/src/ssl/error.rs
```diff
@@ -152,21 +153,42 @@ impl<S: fmt::Debug> StdError for HandshakeError<S> {
...
+            write!(f, "{}: cert verification failed - {}", prefix, verify)?;
+        } else {
+            write!(f, "{}", prefix)?;
+        }
+        if let Some(error) = s.error().io_error() {
```
inikulin, line 169: Style nit: separate multiline expressions and statements by blank lines

inikulin (changes requested):

PR #19, opened 2021-03-15
Tweak yet again the boring error reporting

nox:
We also omit file and line in `ErrorStack` itself now too.

`ErrorStack` is the wrapped error type returned by `hyper_boring::HttpsConnector::call`.

### DIFF
--- a/boring/src/error.rs
+++ b/boring/src/error.rs
@@ -60,16 +60,16 @@ impl ErrorStack {
 impl fmt::Display for ErrorStack {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
         if self.0.is_empty() {
-            return fmt.write_str("OpenSSL error");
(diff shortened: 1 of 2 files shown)


boring/src/error.rs
```diff
@@ -60,16 +60,16 @@ impl ErrorStack {
 impl fmt::Display for ErrorStack {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
         if self.0.is_empty() {
-            return fmt.write_str("OpenSSL error");
+            return fmt.write_str("unknown OpenSSL error");
```
inikulin, line 63: "unknown BoringSSL error"

inikulin (changes requested):

boring/src/error.rs
nox: I changed it to BoringSSL in both places.

inikulin (approved):

PR #21, opened 2021-03-19
Fix wrong assert if padding is disabled

leisim:


inikulin:
@leisim  Could you give more context for the issue that this one fixes, please? Regression test would be useful too

leisim:
@inikulin 
This assert `assert!(output.len() >= input.len() + block_size);` requires the output size to be greater than the input size to make sure there is enough space for padding.
If no padding is used (not necessary if the input size is a multiple of the cipher block size) the output of the encryption has the exact same size as the input. 

I'll add a regression test :+1:

inikulin:
@leisim seems like needs a rebase and an integration test

nox:
@simc Ping?

PR #31, opened 2021-04-16
Introduce HandshakeError::code

nox:


nox:
Oh, that doesn't work as I expected it to. I wonder if we should special case those specific errors during handshake only then.

nox:
I've repurposed the PR to be about exposing the handshake error's code so that we can inspect it on our own from where we consume the damn handshake error.

inikulin (approved):

Issue #36, opened 2021-05-18
Support pre-built binaries of BoringSSL


Evrard-Nil:
Hi all, this is a proposition for an enhancement. 

When using this crate as part of a large project, users may have a build system already building BoringSSL and want to rely on it instead of building it multiple times. This happened to us and thought that others may have the same need.

This feature would allow users to specify a path to the pre-built binaries and one for the corresponding headers in case the version is not the same. It could be done using environment variables which would be checked during the build in boring-sys. Default behavior would stay the same.

jyn514:
Added in https://github.com/cloudflare/boring/pull/37.

PR #37, opened 2021-05-20
Allow pre-built binaries of BoringSSL

Evrard-Nil:
Following this [issue](https://github.com/cloudflare/boring/issues/36).


inikulin:
@Evrard-Nil lgtm, the README also needs to be updated, including instructions of how to compile with pre-built binaries

Evrard-Nil:
@inikulin could you run the workflow please? Also let me know if you feel something is missing from the doc.

cjmakes (approved):

PR #39, opened 2021-06-10
Update bindgen to 0.58.1

roblabla:
Updates bindgen to the latest release.

### DIFF
--- a/boring-sys/Cargo.toml
+++ b/boring-sys/Cargo.toml
@@ -27,5 +27,5 @@ include = [
 ]
 
 [build-dependencies]
-bindgen = "0.57"
+bindgen = "0.58.1"
 cmake = "0.1"


boring-sys/Cargo.toml
```diff
@@ -27,5 +27,5 @@ include = [
 ]
 
 [build-dependencies]
-bindgen = "0.57"
+bindgen = "0.58.1"
```
jyn514, line 30: The latest version is now 0.59. I don't think we need to require later than .0, cargo will install the latest anyway and it gives our dependents more flexibility.
```suggestion
bindgen = "0.59"
```

jyn514 (commented):
Sorry for the long delay. This looks good to me except that bindgen has been updated again in the meantime.

PR #40, opened 2021-06-14
Remove unused dependency on bytes

roblabla:
Bytes is unused by hyper_boring, so remove it from dependency list.


jyn514 (approved):
Looks good to me - do you mind also removing `boring-sys` at the same time? It's not used either.

PR #41, opened 2021-07-27
Note that the build will fail unless you run `git submodule update`

jyn514:
r? @cjmakes


cjmakes (approved):

cjmakes:
Looks like these failures are related to some new lints in rust 1.53 (https://github.com/rust-lang/rust-bindgen/issues/1651). I think these will go away with #39

PR #48, opened 2021-07-30
Pin rust version to 1.54.0

cjmakes:
In order to avoid breaking CI on master with linters added in new rust versions, we should pin to the current version, and updated when there are features we find desirable.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,8 +18,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Install Rust
-        run: rustup update stable && rustup default stable
       - name: Check formatting
         run: cargo fmt --all -- --check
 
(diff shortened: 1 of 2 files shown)


jyn514 (commented):
Can you also update the workflow file? e.g. this bit should be removed: https://github.com/cloudflare/boring/blob/993f68ded2a9a35c52c919c87167400a3971663c/.github/workflows/ci.yml#L22

and I'm not sure how this interacts with explicit toolchains - possibly all of those need to be updated? Maybe add a new `{{version}}` variable? https://github.com/cloudflare/boring/blob/993f68ded2a9a35c52c919c87167400a3971663c/.github/workflows/ci.yml#L83

cjmakes:
> and I'm not sure how this interacts with explicit toolchains

I _think_ we're good here. Pattern matching on wrangler, looks like they do the same.
https://github.com/cloudflare/wrangler/blob/master/.github/workflows/test.yml#L21

nox:
I feel like this is a bad idea and we should instead either accept that sometimes the CI is red for reasons outside our control, or not emit warnings as errors, but that's not my call.

jyn514:
Hmm, my concern about accepting that CI is sometimes red is that can lead to multiple people trying to fix the warnings at once, since everyone making a new PR will see the errors. If we link fixing the errors to updating the toolchain we can avoid duplicating effort.

@nox can you expand more on why you think this is a bad idea?

PR #50, opened 2021-08-12
Replace OUT_DIR variable with BINDGEN_SRC for include! macro

timothyklim:
This small fix can help build this crate with bazel and rules_rust so easily.
Bazel can locate single file with $(locate) but cannot pass a directory of a single file with an absolute path.

### DIFF
--- a/boring-sys/build.rs
+++ b/boring-sys/build.rs
@@ -272,8 +272,11 @@ fn main() {
     }
 
     let bindings = builder.generate().expect("Unable to generate bindings");
-    let out_path = PathBuf::from(env::var("OUT_DIR").unwrap());
+    let out_dir = env::var("OUT_DIR").unwrap();
+    let out_path = PathBuf::from(&out_dir);
     bindings
         .write_to_file(out_path.join("bindings.rs"))
         .expect("Couldn't write bindings!");
+
+    println!("cargo:rustc-env=BINDGEN_SRC={}/bindings.rs", out_dir);
 }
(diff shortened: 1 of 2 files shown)


jyn514:
@timothyklim `OUT_DIR` is pretty standard for rust projects using bindgen - do you make this same PR to all the projects you use with Bazel? It seems strange to me that Bazel would try and change the whole rust ecosystem rather than adapting to make this possible.

timothyklim:
> @timothyklim `OUT_DIR` is pretty standard for rust projects using bindgen - do you make this same PR to all the projects you use with Bazel? It seems strange to me that Bazel would try and change the whole rust ecosystem rather than adapting to make this possible.

To make possible to build this library there more steps:
1. convert entire build.rs to BUILD file without `git clone` execution (becase its not available in the sandbox and there no available internet to make git clone works);
2. concat boringssl headers into single `.h` file for rust bindgen.

>do you make this same PR to all the projects you use with Bazel?
No. For example, crate `openssl-sys` supports pointing to openssl directory without git cloning and example for `openssl-sys` crate includes in rules_rust.

timothyklim:
After some tests and custom rule I can build bindings within out_dir filegroup and successfully pass as env variable. Thanks!

ibigbug:
@timothyklim do you by any chance have a working example of your setup with bazel? I'm somehow facing similar issue that this build.rs expect the static libs to be placed in a particular folder structure which bazel doesn't do natually.

timothyklim:
> @timothyklim do you by any chance have a working example of your setup with bazel? I'm somehow facing similar issue that this build.rs expect the static libs to be placed in a particular folder structure which bazel doesn't do natually.

I've switched to rust-openssl and boringssl#master-with-bazel.

PR #51, opened 2021-08-12
Update `boring` to 2018 edition

jyn514:
I recommend reviewing this by running `cargo fix --edition` locally and making sure the output is the same; I know the diff is giant.

This found a bug in cargo: https://github.com/rust-lang/cargo/issues/9782.


cjmakes (approved):
Matches what my machine says :smile:

PR #52, opened 2021-09-17
Add fips-3678 feature

behrat:
This PR contains multiple logical commits. The ones before "Add fips-3678" I think can actually stand on their own, and are useful even without FIPS support.

Since some of the APIs are not available in the FIPS-validated version, we must add some conditional-compilation in order to successfully link against the FIPS-validated version of BoringSSL (see the addition in `README.md` for details). This can be controlled with the new `fips-3678` feature. The intention of naming the feature this way is that future FIPS-validated versions of BoringSSL can by supported by adding a `fips-xxxx` feature, while still supporting the `fips-3678` for backwards-compatability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,20 +27,11 @@ _Warning_: When providing a different version of BoringSSL make sure to use a co
 Only BoringCrypto module version ae223d6138807a13006342edfeef32e813246b39, as
 certified with [certificate
 3678](https://csrc.nist.gov/projects/cryptographic-module-validation-program/certificate/3678)
-is supported by this crate. Support is enabled by this crate's `fips-3678` feature.
+is supported by this crate. Support is enabled by this crate's `fips` feature.
 
-You must provide and build your own FIPS-validated module in order for this
-crate to link against it (see the instalation instructions in the
-[BoringCrypto FIPS 140-2 Non-Proprietary Security Policy] (https://csrc.nist.gov/CSRC/media/projects/cryptographic-module-validation-program/documents/security-policies/140sp3678.pdf)
-).
-
-After building the FIPS-validated module, you must build this crate with the
-`BORING_BSSL_PATH`, `BORING_BSSL_INCLUDE_PATH`, and maybe the
-`BORING_BSSL_LIB_PATH`. Once built, the `fips_enabled` example can be used to
-test that `boring` is running in FIPS mode:
+`boring-sys` comes with a test that FIPS is enabled/disabled depending on the feature flag. You can run it as follows:
 ```bash
-$ cargo run --features fips-3678 --example fips_enabled
-boring::fips::enabled(): true
+$ cargo test --features fips fips::is_enabled
 ```
 
 ## Contribution
--- a/README.md
+++ b/README.md
@@ -22,6 +22,27 @@ _Notes_: The crate will look for headers in the `$BORING_BSSL_INCLUDE_PATH/opens
 
 _Warning_: When providing a different version of BoringSSL make sure to use a compatible one, the crate relies on the presence of certain functions.
 
+## Building with a FIPS-validated module
+
+Only BoringCrypto module version ae223d6138807a13006342edfeef32e813246b39, as
+certified with [certificate
+3678](https://csrc.nist.gov/projects/cryptographic-module-validation-program/certificate/3678)
+is supported by this crate. Support is enabled by this crate's `fips-3678` feature.
+
+You must provide and build your own FIPS-validated module in order for this
+crate to link against it (see the instalation instructions in the
+[BoringCrypto FIPS 140-2 Non-Proprietary Security Policy] (https://csrc.nist.gov/CSRC/media/projects/cryptographic-module-validation-program/documents/security-policies/140sp3678.pdf)
+).
+
+After building the FIPS-validated module, you must build this crate with the
+`BORING_BSSL_PATH`, `BORING_BSSL_INCLUDE_PATH`, and maybe the
+`BORING_BSSL_LIB_PATH`. Once built, the `fips_enabled` example can be used to
+test that `boring` is running in FIPS mode:
+```bash
+$ cargo run --features fips-3678 --example fips_enabled
+boring::fips::enabled(): true
+```
+
 ## Contribution
 
 Unless you explicitly state otherwise, any contribution intentionally
--- a/boring-sys/build.rs
+++ b/boring-sys/build.rs
@@ -291,11 +291,14 @@ fn main() {
         println!("cargo:rustc-cdylib-link-arg=-Wl,-undefined,dynamic_lookup");
     }
 
-    let include_path = if cfg!(feature = "fips") {
-        format!("{}/include", BORING_SSL_PATH)
-    } else {
-        format!("{}/src/include", BORING_SSL_PATH)
-    };
+    println!("cargo:rerun-if-env-changed=BORING_BSSL_INCLUDE_PATH");
+    let include_path = std::env::var("BORING_BSSL_INCLUDE_PATH").unwrap_or_else(|_| {
+        if cfg!(feature = "fips") {
+            format!("{}/include", BORING_SSL_PATH)
+        } else {
+            format!("{}/src/include", BORING_SSL_PATH)
+        }
+    });
 
     let mut builder = bindgen::Builder::default()
         .derive_copy(true)
--- a/boring-sys/build.rs
+++ b/boring-sys/build.rs
@@ -176,25 +176,25 @@ fn main() {
     use std::path::{Path, PathBuf};
     use std::process::Command;
 
-    if !Path::new("deps/boringssl/CMakeLists.txt").exists() {
-        println!("cargo:warning=fetching boringssl git submodule");
-        // fetch the boringssl submodule
-        let status = Command::new("git")
-            .args(&[
-                "submodule",
-                "update",
-                "--init",
-                "--recursive",
-                "deps/boringssl",
-            ])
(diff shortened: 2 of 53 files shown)


boring-sys/build.rs
```diff
@@ -176,24 +176,25 @@ fn main() {
...
-                "submodule",
-                "update",
-                "--init",
-                "--recursive",
-                "deps/boringssl",
-            ])
-            .status();
-        if !status.map_or(false, |status| status.success()) {
-            panic!("failed to fetch submodule - consider running `git submodule update --init --recursive deps/boringssl` yourself");
+    println!("cargo:rerun-if-env-changed=BORING_BSSL_PATH");
```
jyn514, line 179: Hmm, shouldn't we have this for all the other env variables too? Off the top of my head I see `OPT_LEVEL`, `DEBUG`, `CARGO_CFG_TARGET_ARCH`, `CARGO_CFG_TARGET_OS`, ... https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-env-changed says that env variables set by cargo are automatically rerun, but I suspect there's still a few that aren't being caught.

Happy to do that in a separate PR to avoid making this one too big.

boring/src/x509/mod.rs
```diff
@@ -493,18 +493,18 @@ impl X509Ref {
     /// Returns the certificate's Not After validity period.
     pub fn not_after(&self) -> &Asn1TimeRef {
         unsafe {
-            let date = X509_getm_notAfter(self.as_ptr());
+            let date = X509_get0_notAfter(self.as_ptr());
```
jyn514, line 511: > since we're returning an unmutable reference, it's not possible to call any mutable functions of Asn1TimeRef (without using unsafe again): https://docs.rs/boring/1.1.6/boring/asn1/struct.Asn1TimeRef.html Another way to say it: returning an unmutable reference is effectively propagating the same guarantee as *const. (right?)

Doesn't an immutable reference guarantee that it won't be changed by this code *or any other code*? The docs say

> X509_getm_notBefore() and X509_getm_notAfter() are similar to X509_get0_notBefore() and X509_get0_notAfter() except they return non-constant mutable references

which seems to imply that these could change because the OpenSSL library itself decides to change them. I don't know if that's valid while another piece of code has an immutable reference to them.

boring-sys/build.rs
```diff
@@ -205,9 +205,12 @@ fn main() {
         cfg.build_target("bssl").build().display().to_string()
     });
 
-    let build_path = get_boringssl_platform_output_path();
-    let build_dir = format!("{}/build/{}", bssl_dir, build_path);
-    println!("cargo:rustc-link-search=native={}", build_dir);
+    println!("cargo:rerun-if-env-changed=BORING_BSSL_LIB_PATH");
+    let lib_path = std::env::var("BORING_BSSL_LIB_PATH").unwrap_or_else(|_| {
```
jyn514, line 209: Why this change? What are you using it for?

README.md
```diff
@@ -22,6 +22,27 @@ _Notes_: The crate will look for headers in the `$BORING_BSSL_INCLUDE_PATH/opens
...
+3678](https://csrc.nist.gov/projects/cryptographic-module-validation-program/certificate/3678)
+is supported by this crate. Support is enabled by this crate's `fips-3678` feature.
+
+You must provide and build your own FIPS-validated module in order for this
+crate to link against it (see the instalation instructions in the
+[BoringCrypto FIPS 140-2 Non-Proprietary Security Policy] (https://csrc.nist.gov/CSRC/media/projects/cryptographic-module-validation-program/documents/security-policies/140sp3678.pdf)
+).
+
+After building the FIPS-validated module, you must build this crate with the
+`BORING_BSSL_PATH`, `BORING_BSSL_INCLUDE_PATH`, and maybe the
```
jyn514, line 38: Why is this true? What happens if you don't set these variables?

boring/src/x509/mod.rs
```diff
@@ -229,16 +228,28 @@ impl X509Builder {
...
     pub fn set_not_after(&mut self, not_after: &Asn1TimeRef) -> Result<(), ErrorStack> {
-        unsafe { cvt(X509_set1_notAfter(self.0.as_ptr(), not_after.as_ptr())).map(|_| ()) }
+        #[cfg(feature = "fips")]
+        unsafe {
+            cvt(X509_set_notAfter(self.0.as_ptr(), not_after.as_ptr())).map(|_| ())
+        }
+        #[cfg(not(feature = "fips"))]
+        unsafe {
+            cvt(X509_set1_notAfter(self.0.as_ptr(), not_after.as_ptr())).map(|_| ())
+        }
```
jyn514, line 240: Is `set_notAfter` still present in the latest version? Can we use that unconditionally instead of adding two different paths?

jyn514 (changes requested):
Can you add a test to `.github/workflows/ci.yml` testing that this compiles when `--features fips-3678` is set?

boring-sys/build.rs
behrat: In my build environment, the I must set the include path and library path to different values, or example:
```
$ env | grep BORING                                  
BORING_BSSL_INCLUDE_PATH=/opt/boringssl-fips/include
BORING_BSSL_LIB_PATH=/opt/boringssl-fips
```
Only then will it successfully build and link with my FIPS build of BoringSSL:
```
$ cargo run --features fips-3678 --example fips_enabled
   Compiling boring-sys v1.1.1 (/home/braden/rust/boring/boring-sys)
   Compiling boring v1.1.6 (/home/braden/rust/boring/boring)
    Finished dev [unoptimized + debuginfo] target(s) in 18.00s
     Running `/home/braden/rust/boring/target/debug/examples/fips_enabled`
boring::fips::enabled(): true
```
Open to any suggestions to simply things or make it work better in other use cases, though.

boring/src/x509/mod.rs
jyn514: I think so, yes - that way the same code gets tested whether the feature is enabled or disabled. Maybe add a comment that this should be changed when the version of FIPS is bumped (I'm not sure what the process for that is; I guess the C boring lib itself has to get recertified?).

README.md
jyn514: Hmm, ideally we could somehow give a compile error if you try and use a non-FIPS library when the FIPS feature is enabled. Maybe we can have it link to a symbol that only exists in the FIPS version? Or add an assert somewhere?

boring-sys/build.rs
jyn514: :+1: that would be great, thanks!

boring/src/x509/mod.rs
jyn514: I think I was confused before - those docs are talking about `X509_getm_notAfter`, not `X509_get0_notAfter`. I think get0 should be ok as you're written it.

jyn514:
I opened https://github.com/behrat/boring/pull/1 getting this working a little better without needing an external boring build.

boring-sys/build.rs
```diff
@@ -216,10 +291,11 @@ fn main() {
...
 
-    let include_path = PathBuf::from(
-        std::env::var("BORING_BSSL_INCLUDE_PATH")
-            .unwrap_or_else(|_| String::from("deps/boringssl/src/include")),
-    );
+    let include_path = if cfg!(feature = "fips") {
+        format!("{}/include", BORING_SSL_PATH)
+    } else {
+        format!("{}/src/include", BORING_SSL_PATH)
+    };
```
jyn514, line 298: I think I was confused before, `BORING_BSSL_INCLUDE_PATH` was supported even before this PR. The new code should look something like this:
```rust
    println!("cargo:rerun-if-env-changed=BORING_BSSL_INCLUDE_PATH");
    let include_path = std::env::var("BORING_BSSL_INCLUDE_PATH").unwrap_or_else(|| if cfg!(feature = "fips") {
        format!("{}/include", BORING_SSL_PATH)
    } else {
        format!("{}/src/include", BORING_SSL_PATH)
    });
```
Do you want to make that change or should I?

jyn514 (commented):
Looks good with the `BORING_BSSL_INCLUDE_PATH` support added back in - sorry for the noise there.

jyn514:
Oh, also it looks like you need to add clang-7 to the test environment.

boring/Cargo.toml
```diff
@@ -21,3 +21,8 @@ boring-sys = { version = "1.1.0", path = "../boring-sys" }
 [dev-dependencies]
 hex = "0.4"
 rusty-hook = "^0.11"
+
+[features]
+fips-3678 = ["fips", "boring-sys/fips-3678"]
+# The "fips" feature is used for conditional-compilation that is common to all supported FIPS-validated versions
+fips = ["boring-sys/fips"]
```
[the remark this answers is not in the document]
behrat: Discussed in https://github.com/behrat/boring/pull/1 and decided to only have `fips` feature for now (could support multiple versions later).

boring-sys/build.rs
behrat: Do you mind making the change? I promise to review quickly. :)

jyn514: Pushed in https://github.com/cloudflare/boring/pull/52/commits/0e742156f58dc6ff4846efaefb4f95c17c5549ab.

behrat:
Thanks for the help @jyn514! I think that addresses all feedback from my original version of the PR, and this looks ready-to-merge to me.

wmorgan:
Thanks for this @jyn514 and @behrat!

PR #53, opened 2021-10-06
Remove PasswordCallback from boring-sys

deeprobin:
I have removed the PasswordCallback. References to this are not present in the repository, but is it ok to delete this directly or should this be marked as `#[deprecated]` first?

Best regards
Robin


jyn514:
Thanks for the PR! Sorry for the long delay.

PR #55, opened 2021-11-02
boring-sys: Disable unnecessary bindgen dependencies

olix0r:
`bindgen`'s default features include dependencies like `clap` and
`env_logger` which are unnecessary when building `boring-sys`. When
using tools like `cargo-deny` to audit dependencies with mismatched
versions, these dependencies can easily conflict with application
dependencies.

This change disables `bindgen`'s default features to avoid unnecessary
dependencies.

### DIFF
--- a/boring-sys/Cargo.toml
+++ b/boring-sys/Cargo.toml
@@ -27,5 +27,5 @@ include = [
 ]
 
 [build-dependencies]
-bindgen = "0.59"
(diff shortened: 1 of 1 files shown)


jyn514 (approved):
Looks great, thanks!

olix0r:
@jyn514 Thanks for the quick review :)

PR #56, opened 2021-11-04
Relax constraints on `Display for tokio-boring::HandshakeError`

olix0r:
`tokio_boring::HandshakeError` currently requires that the inner stream
implements `Debug` for its `Display` implementation. This constraint is
unnecessary.

This change removes this `Debug` constraint so `HandshakeError` always
implements `Display`.


jyn514 (approved):
Thanks!

Issue #20, opened 2021-03-17
The error types should be refactored


nox:
I have many issues with the various `Error` types we define and how `HttpsConnector` ultimately just uses `BoxError` for its `Service<Uri>` error type, I'll try to summarize them here.

First, the `BoxError`, this makes it impossible to consume any more specific error type, as downcasting with the `Error` trait is always by reference.

Second, I keep confusing myself with `boring::Error` and `boring::ssl::Error`.

Third, the `boring::ssl::HandshakeError` is not fun to use for multiple reasons:
 * the underlying I/O errors can be in two separate variants (`Failure` and `WouldBlock`);
 * it stores a `MidHandshakeSslStream<S>` even in the `Failure` variant, even though you are obviously not supposed to do anything anymore with that stream given the handshake failed;
* given that second bullet point, it also doesn't make much sense for the `MidHandshakeSslStream<S>` struct to keep around the error that interrupted the handshake, as that was expected and you just want to resume it;
 * it has a variant `SetupFailure` which feels out of place to me, shouldn't setup errors be completely contained in builders etc?

Fourth, `tokio_boring::HandshakeError` is as useful as `BoxError` given it doesn't let us access the `boring::ssl::HandshakeError` it wraps directly, so that's one more layer of hoops to go through to find, say, I/O errors.

Fifth, even if `tokio_boring::HandshakeError` let us access its inner `boring::ssl::HandshakeError`, that would still be a bit of a bother to use, as we *know* `tokio_boring` would never return a `WouldBlock` error but we would still need an arm for that in our code.

inikulin:
@nox let's do these improvements

nox:
As a first step, I want to change:

```rust
pub enum HandshakeError<S> {
    /// Setup failed.
    SetupFailure(ErrorStack),
    /// The handshake failed.
    Failure(MidHandshakeSslStream<S>),
    /// The handshake encountered a `WouldBlock` error midway through.
    ///
    /// This error will never be returned for blocking streams.
    WouldBlock(MidHandshakeSslStream<S>),
}

pub struct MidHandshakeSslStream<S> {
    stream: SslStream<S>,
    error: Error,
}
```

to

```rust
pub enum HandshakeResult<S> {
    Done(SslStream<S>),
    WouldBlock(MidHandshakeSslStream<S>),
    Failed(StreamError<S>),
}

pub struct StreamError<S> {
    stream: S,
    error: Error,
}

pub struct MidHandshakeSslStream<S> {
[...]

PR #57, opened 2021-11-04
tokio-boring: Add additional accessors to `HandshakeError`

hawkw:
Currently, the `HandshakeError` type in `tokio-boring` provides a
`std::error::Error` implementation and an `as_io_error` method that
returns an `Option<&std::io::Error>`, as the only ways to access the
potentially underlying error. There are a couple of cases where this is
somewhat insufficient:

* If a user has an owned `HandshakeError`, and needs an owned
  `io::Error`, they have to create a *new* one, using the
  `HandshakeError`'s cause and message, like this:
  ```rust
  let err: HandshakeError<_> = // ...;
  if let Some(io_err) = err.as_io_error() {
      return Err(io::Error::new(io_err.cause(), io_err.to_string()));
  }
  // ...
  ```
  This seems like kind of a shame, since it allocates two additional
  times (for the `String` and then again for the `io::Error` itself),
  and deallocates the original `io::Error` at the end of the scope. None
  of this *should* be necessary, since the `HandshakeError` in this case
  is owned and the original `io::Error` could be returned.

* `HandshakeError` only implements `fmt::Debug` and `fmt::Display` when
  the wrapped I/O stream it's generic over implements `Debug`. This
  means that bounding the `S` type with `Debug` is necessary for
  `HandshakeError` to implement the `Error` trait. In some cases,
  introducing a `Debug` bound on a generic I/O type is kind of a
  heavyweight requirement. Therefore, it would be nice to have a way to
  extract a type that *always* implements `Error` from the
  `HandshakeError`, in cases where the returned I/O stream is just going
  to be discarded if the handshake failed.

This branch introduces new functions on `HandshakeError`:

* `HandshakeError::as_ssl_error`, which is analogous to
  `HandshakeError::as_io_error` but returning an
  `Option<&boring::error::ErrorStack>` instead of an I/O error.
* `HandshakeError::into_io_error` and `HandshakeError::into_ssl_error`,
  which consume the error and return an `Option<io::Error>` or an
  `Option<ErrorStack>`, respectively.

  I noticed that some of the `into_$TYPE` methods on errors in the
  `boring` crate return `Result`s rather than `Option`s so that the
  original error can be reused if it couldn't be converted into the
  requested type. However, this can't easily be done in
  `HandshakeError`'s `into_io_error` and `into_ssl_error`, because these
  methods call `MidHandshakeSslStream::into_error`, and (since
  `MidHandshakeSslStream` can only be constructed by the `boring` crate,
  not in `tokio-boring`), we can't ever get the `MidHandshakeSslStream`
  _back_...so we can't return the original `Self` type in this case.

Hopefully this should make it much easier to convert `HandshakeError`s
into other error types as required in user code!

Signed-off-by: Eliza Weisman <eliza@buoyant.io>

### DIFF
--- a/tokio-boring/src/lib.rs
+++ b/tokio-boring/src/lib.rs
@@ -307,6 +307,40 @@ impl<S> HandshakeError<S> {
             _ => None,
         }
     }
+
+    /// Returns a reference to the inner SSL error, if this error was caused by
+    /// a SSL error.
+    pub fn as_ssl_error(&self) -> Option<&boring::error::ErrorStack> {
+        match &self.0 {
+            ssl::HandshakeError::Failure(s) => s.error().ssl_error(),
+            ssl::HandshakeError::SetupFailure(ref s) => Some(s),
+            _ => None,
+        }
+    }
+
+    /// Consumes `self` and returns the inner I/O error by value, if this error
+    /// was caused by an I/O error.
+    pub fn into_io_error(self) -> Option<io::Error> {
+        match self.0 {
+            ssl::HandshakeError::Failure(s) => s.into_error().into_io_error().ok(),
+            _ => None,
+        }
+    }
+
+    /// Consumes `self` and returns the inner SSL error by value, if this error
+    /// was caused by n SSL error.
+    pub fn into_ssl_error(self) -> Option<boring::error::ErrorStack> {
+        match self.0 {
+            // TODO(eliza): it's not great that we have to clone in this case,
+            // but `boring::ssl::Error` doesn't currently expose an
+            // `into_ssl_error` the way it does for `io::Error`s. We could add
+            // that in a separate PR, but adding that here would make
+            // `tokio-boring` depend on a new release of `boring`...
+            ssl::HandshakeError::Failure(s) => s.into_error().ssl_error().cloned(),
+            ssl::HandshakeError::SetupFailure(stack) => Some(stack),
+            _ => None,
+        }
+    }
 }
 
 impl<S> fmt::Debug for HandshakeError<S>


tokio-boring/src/lib.rs
```diff
@@ -307,6 +307,40 @@ impl<S> HandshakeError<S> {
...
+        match &self.0 {
+            ssl::HandshakeError::Failure(s) => s.error().ssl_error(),
+            ssl::HandshakeError::SetupFailure(ref s) => Some(s),
+            _ => None,
+        }
+    }
+
+    /// Consumes `self` and returns the inner I/O error by value, if this error
+    /// was caused by an I/O error.
+    pub fn into_io_error(self) -> Option<io::Error> {
```
olix0r, line 323: FWIW I'm not sure that this is valuable for us... If it's not an Io error we lose the original error?

I guess we'd have to use something like `if e.as_io_error().is_some() { e.into_io_error.unwrap() {` ? that's pretty cumbersome

hawkw: I thought about adding an `is_io`/`is_ssl`, but I wasn't sure if it was necessary given that it would just be `as_io_error().is_some()`. I could add that as well, though. IMO it's still less cumbersome than the current code we have, which calls `as_io_error` and creates a new `io::Error` using values from the one that's passed by reference.

Unfortunately, we can't

hawkw: @olix0r okay, I've modified these to return `Result<io::Error, Self>` and `Result<ErrorStack, Self>` so you can now write code like
```rust
error
    .into_ssl_error()
    .map(Into::into)
    .or_else(|error| error.into_io_error().map(Into::into))
```
or similar

olix0r: awesome!

tokio-boring/src/lib.rs
```diff
@@ -307,6 +307,58 @@ impl<S> HandshakeError<S> {
...
         }
     }
+
+    /// Returns a reference to the inner SSL error, if this error was caused by
+    /// a SSL error.
+    pub fn as_ssl_error(&self) -> Option<&boring::error::ErrorStack> {
+        match &self.0 {
+            ssl::HandshakeError::Failure(s) => s.error().ssl_error(),
+            ssl::HandshakeError::SetupFailure(ref s) => Some(s),
+            _ => None,
```
jyn514, line 317: Shouldn't this also handle `HandshakeError::WouldBlock`? Or is it impossible somehow for WouldBlock to be an SSL error?
```suggestion
            ssl::HandshakeError::Failure(s) | ssl::HandshakeError::WouldBlock(s) => s.error().ssl_error(),
            ssl::HandshakeError::SetupFailure(ref s) => Some(s),
```

tokio-boring/src/lib.rs
```diff
@@ -307,6 +307,58 @@ impl<S> HandshakeError<S> {
...
+    }
+
+    /// Consumes `self` and returns the inner I/O error by value, if this error
+    /// was caused by an I/O error.
+    ///
+    /// If this error was not caused by an I/O error, returns `Err(self)` so
+    /// that the error value can be reused.
+    pub fn into_io_error(self) -> Result<io::Error, Self> {
+        match self.0 {
+            ssl::HandshakeError::Failure(s) => {
```
jyn514, line 328:
```suggestion
            ssl::HandshakeError::Failure(s) | ssl::HandshakeError::WouldBlock(s) => {
```

tokio-boring/src/lib.rs
```diff
@@ -307,6 +307,58 @@ impl<S> HandshakeError<S> {
...
+                if s.error().io_error().is_some() {
+                    Ok(s.into_error().into_io_error().expect(
+                        "if `s.error().io_error().is_some()`, `into_io_error()` must succeed",
+                    ))
+                } else {
+                    Err(Self(ssl::HandshakeError::Failure(s)))
+                }
+            }
+            _ => Err(self),
```
jyn514, line 337:
```suggestion
            ssl::HandshakeError::SetupFailure(_) => Err(self),
```

tokio-boring/src/lib.rs
```diff
@@ -307,6 +307,58 @@ impl<S> HandshakeError<S> {
...
+            // TODO(eliza): it's not great that we have to clone in this case,
+            // but `boring::ssl::Error` doesn't currently expose an
+            // `into_ssl_error` the way it does for `io::Error`s. We could add
+            // that in a separate PR, but adding that here would make
+            // `tokio-boring` depend on a new release of `boring`...
+            ssl::HandshakeError::Failure(s) => s
+                .error()
+                .ssl_error()
+                .cloned()
+                .ok_or_else(|| Self(ssl::HandshakeError::Failure(s))),
```
jyn514, line 357: I'd rather add `into_ssl_error` for boring too, if you don't mind - you can change them both in the same PR.

tokio-boring/src/lib.rs
```diff
@@ -307,6 +307,58 @@ impl<S> HandshakeError<S> {
...
+    ///
+    /// If this error was not caused by an I/O error, returns `Err(self)` so
+    /// that the error value can be reused.
+    pub fn into_io_error(self) -> Result<io::Error, Self> {
+        match self.0 {
+            ssl::HandshakeError::Failure(s) => {
+                if s.error().io_error().is_some() {
+                    Ok(s.into_error().into_io_error().expect(
+                        "if `s.error().io_error().is_some()`, `into_io_error()` must succeed",
+                    ))
```
jyn514, line 332: I think you could avoid the `expect()` here by adding a new `MidHandshakeSslStream::from_parts` function and using `into_parts()` above.

jyn514 (commented):
Sorry for the long delay. It makes sense to me that the existing API is insufficient, and I think the type signatures for the new functions are fine. Do be aware though that we're considering larger changes to the error types soon: https://github.com/cloudflare/boring/issues/20

tokio-boring/src/lib.rs
hawkw: Sure! I was trying to limit the API additions because I wasn't sure what would be preferred by the maintainers, but I'm happy to change that.

nox:
Btw #142 completely removes `HandshakeError`.

Issue #54, opened 2021-10-08
New boring-sys release


vthib:
Would it be possible to release a new version of boring-sys in the near future? I would like to be able to use the bindgen dependency update that got merged a few month ago.

olix0r:
I could also use a new release of `boring-sys`, but it would be great if https://github.com/cloudflare/boring/pull/55 could be included in it :)

PR #59, opened 2021-12-16
Bump versions for release

jyn514:
Closes https://github.com/cloudflare/boring/issues/54

- Major version for boring-sys: `PasswordCallback` was removed
- Major version for boring: the public `*Ref` types were removed and `foreign-types` appears in our public api and had a major version bump
- Patch version for tokio-boring: the only API change was removing the `S: Debug` bound
- Patch version for hyper-boring: no API changes, only removed dependencies

### DIFF
--- a/hyper-boring/Cargo.toml
+++ b/hyper-boring/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hyper-boring"
-version = "2.1.0"
+version = "2.1.1"
 authors = ["Steven Fackler <sfackler@gmail.com>", "Ivan Nikulin <ifaaan@gmail.com>"]
 edition = "2018"
 description = "Hyper TLS support via BoringSSL"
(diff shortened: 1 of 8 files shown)


hyper-boring/Cargo.toml
```diff
@@ -21,7 +21,7 @@ http = "0.2"
 hyper = { version = "0.14", default-features = false, features = ["client"] }
 linked_hash_set = "0.1"
 once_cell = "1.0"
-boring = { version = "1.1.0", path = "../boring" }
+boring = { version = "2", path = "../boring" }
```
inikulin, line 24: I think we need a major bump for those too: considering that you can't use 2 versions of boring side by side, this implies that in order to use this version of `hyper-boring` you would need to bump boring dep for your app, otherwise it won't compile. So, it's technically a breaking change too.

inikulin (changes requested):

hyper-boring/Cargo.toml
jyn514: Oh - we can fix that by allowing either the old or new version of `boring` I think. Let me look up the semver constraint for that.

jyn514: Ok, done.

inikulin (approved):

Issue #22, opened 2021-03-20
Error building iOS binary


simc:
I'm getting the following error when I try to build `boring` for iOS. Do you have an idea what went wrong?

<details><summary>Error</summary>
<p>

```
...
Compiling boring-sys v1.1.0
error: failed to run custom build command for `boring-sys v1.1.0`

Caused by:
  process didn't exit successfully: `/Users/runner/work/isar-core/isar-core/dart-ffi/target/release/build/boring-sys-a0cbecad694041d4/build-script-build` (exit code: 101)
  --- stdout
  running: "xcrun" "--show-sdk-path" "--sdk" "iphoneos"
  exit code: 0
  running: "xcrun" "--show-sdk-path" "--sdk" "iphoneos"
  exit code: 0
  running: "cmake" "/Users/runner/.cargo/git/checkouts/boring-dfc6e3788abbaf96/80c04df/boring-sys/deps/boringssl" "-DCMAKE_OSX_ARCHITECTURES=arm64" "-DCMAKE_OSX_SYSROOT=iphoneos" "-DCMAKE_ASM_FLAGS=-fembed-bitcode " "-DCMAKE_INSTALL_PREFIX=/Users/runner/work/isar-core/isar-core/dart-ffi/target/aarch64-apple-ios/release/build/boring-sys-9d553a3ff05eb868/out" "-DCMAKE_C_FLAGS= -fembed-bitcode  -fPIC --target=aarch64-apple-ios -arch arm64 -miphoneos-version-min=7.0 -isysroot /Applications/Xcode_12.4.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS14.4.sdk -fembed-bitcode" "-DCMAKE_C_COMPILER=/usr/bin/clang" "-DCMAKE_CXX_FLAGS= -fPIC --target=aarch64-apple-ios -arch arm64 -miphoneos-version-min=7.0 -isysroot /Applications/Xcode_12.4.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS14.4.sdk -fembed-bitcode" "-DCMAKE_CXX_COMPILER=/usr/bin/clang++" "-DCMAKE_ASM_COMPILER=/usr/bin/clang" "-DCMAKE_BUILD_TYPE=Release"
  -- The C compiler identification is AppleClang 12.0.0.12000032
  -- The CXX compiler identification is AppleClang 12.0.0.12000032
[...]

PR #60, opened 2021-12-30
Support cross-compilation for iOS

ilammy:
There was some prior work for iOS support, but it did not really produce a successful build because running `bindgen` failed on macOS (#22). That's because bindgen tried using macOS headers instead of iOS ones, and those don't really match. Proper `-isysroot` flag needs to be passed to clang invocations made by bindgen, pointing it into the right direction.

- Fix building for all iOS targets that rustc supports:
  - `aarch64-apple-ios` – iOS devices running ARM64
  - `aarch64-apple-ios-sim` – iOS simulator running on Apple M1
  - `x86_64-apple-ios` – iOS simulator running on x86_64
- Teach CI to test building for aforementioned targets
  -  Only `cargo build`, no `cargo test`, since that's *way* more involved.
- Suppress some Clippy warnings so that the build gets green

I don't have M1 hardware so I have no idea whether `aarch64-apple-ios-sim` target actually works. Cargo says it builds though. Testing would be much appreciated.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -152,11 +152,11 @@ jobs:
     - name: Set LIBCLANG_PATH
       if: startsWith(matrix.os, 'windows')
       run: echo "LIBCLANG_PATH=$((gcm clang).source -replace "clang.exe")" >> $env:GITHUB_ENV
-    - if: startsWith(matrix.os, 'windows')
+    - if: "startsWith(matrix.os, 'windows') && !contains(matrix.target, 'ios')"
       # CI's Windows doesn't have require root certs
       run: cargo test --workspace --exclude tokio-boring --exclude hyper-boring
       name: Run tests (Windows)
-    - if: "!startsWith(matrix.os, 'windows')"
+    - if: "!startsWith(matrix.os, 'windows') && !contains(matrix.target, 'ios')"
       run: cargo test
       name: Run tests (not Windows)
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,6 +70,9 @@ jobs:
         - arm64-android
         - i686-android
         - x86_64-android
+        - aarch64-ios
+        - aarch64-ios-sim
+        - x86_64-ios
         - i686-linux
         - arm-linux
         - aarch64-linux
@@ -102,6 +105,15 @@ jobs:
           target: x86_64-linux-android
           rust: stable
           os: ubuntu-latest
+        - thing: aarch64-ios
+          target: aarch64-apple-ios
+          os: macos-latest
+        - thing: aarch64-ios-sim
+          target: aarch64-apple-ios-sim
+          os: macos-latest
+        - thing: x86_64-ios
+          target: x86_64-apple-ios
+          os: macos-latest
         - thing: i686-linux
           target: i686-unknown-linux-gnu
           rust: stable
@@ -159,6 +171,11 @@ jobs:
     - if: "!startsWith(matrix.os, 'windows') && !contains(matrix.target, 'ios')"
       run: cargo test
       name: Run tests (not Windows)
+    - if: "contains(matrix.target, 'ios')"
+      # It's... theoretically possible to run tests on iPhone Simulator,
+      # but for now, make sure that BoringSSL only builds.
+      run: cargo check --target ${{ matrix.target }} --all-targets
+      name: Check tests (iOS)
 
   test-fips:
     name: Test FIPS integration
--- a/boring-sys/build.rs
+++ b/boring-sys/build.rs
@@ -354,7 +354,8 @@ fn main() {
     println!("cargo:rustc-link-lib=static=ssl");
 
     // MacOS: Allow cdylib to link with undefined symbols
-    if cfg!(target_os = "macos") {
+    let target_os = std::env::var("CARGO_CFG_TARGET_OS").unwrap();
+    if target_os == "macos" {
         println!("cargo:rustc-cdylib-link-arg=-Wl,-undefined,dynamic_lookup");
     }
 
--- a/boring-sys/build.rs
+++ b/boring-sys/build.rs
@@ -82,6 +82,16 @@ fn cmake_params_ios() -> &'static [(&'static str, &'static str)] {
     &[]
 }
 
+fn get_ios_sdk_name() -> &'static str {
+    for (name, value) in cmake_params_ios() {
+        if *name == "CMAKE_OSX_SYSROOT" {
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -152,10 +152,10 @@ jobs:
     - name: Set LIBCLANG_PATH
       if: startsWith(matrix.os, 'windows')
       run: echo "LIBCLANG_PATH=$((gcm clang).source -replace "clang.exe")" >> $env:GITHUB_ENV
-    - if: startsWith(matrix.os, 'windows')
+    - if: "startsWith(matrix.os, 'windows') && !contains(matrix.target, 'ios')"
       # CI's Windows doesn't have require root certs
       run: cargo test --workspace --exclude tokio-boring --exclude hyper-boring
       name: Run tests (Windows)
-    - if: "!startsWith(matrix.os, 'windows')"
+    - if: "!startsWith(matrix.os, 'windows') && !contains(matrix.target, 'ios')"
```
jyn514, line 171: Why this change? If it's because the tests were failing, why didn't they fail before this PR?

.github/workflows/ci.yml
```diff
@@ -159,3 +171,8 @@ jobs:
     - if: "!startsWith(matrix.os, 'windows') && !contains(matrix.target, 'ios')"
       run: cargo test
       name: Run tests (not Windows)
+    - if: "contains(matrix.target, 'ios')"
+      # It's... theoretically possible to run tests on iPhone Simulator,
+      # but for now, make sure that BoringSSL only builds.
+      run: cargo build --target ${{ matrix.target }}
```
jyn514, line 177: This is `build` instead of check to make sure the linker succeeds?

boring/src/pkey.rs
```diff
@@ -93,20 +93,23 @@ impl Id {
...
 /// A trait indicating that a key has public components.
+#[allow(clippy::missing_safety_doc)]
 pub unsafe trait HasPublic {}
 
 unsafe impl HasPublic for Public {}
 
 unsafe impl<T> HasPublic for T where T: HasPrivate {}
 
 /// A trait indicating that a key has private components.
+#[allow(clippy::missing_safety_doc)]
```
jyn514, line 112: Please move this into a separate PR, it's not related to iOS support.

jyn514 (commented):
This seems reasonable, but I'm a little worried that I have no way to test it ... ideally there would be a way to cross-compile from Linux, but I know Apple makes that hard - I know some people at cloudflare with Macs, could you post instructions for how to build for iOS using a Mac?

ilammy:
> ideally there would be a way to cross-compile from Linux, but I know Apple makes that hard

It's theoretically possible – I know people who did this some time ago with C code – but it's technically tedious (you gotta have Xcode SDKs, ideally the same clang, and *ideally* code signing setup). IIRC, it used to be legally dubious too, but nowadays Apple might be cool even if you produce binaries on non-Apple hardware. But however challenging this is, nobody publishing actual apps for iOS would go through this ordeal.

> I know some people at cloudflare with Macs, could you post instructions for how to build for iOS using a Mac?

Since `boring` is just a Rust library, it's a bit involved: you need to build a `staticlib` exporting C API for iOS app to actually use. The build step itself is simply `cargo build --target <suitable-target>`.

Here's the sample that I'm using:

https://github.com/ilammy/boring-example

That's a stub that calls SHA-256 via `boring`.

.github/workflows/ci.yml
ilammy: It's not possible to run `cargo test` when building for iOS. It would produce test binaries, but they can't be executed on the host macOS. You'd need to start an iOS Simulator, wrap the binaries into iOS apps, upload the apps to the simulator, run them there, extract results, and post those back. That's too much hassle. I believe it's okay to skip this step until Rust ecosystem catches up.

So for iOS builds the tests are skipped, only checking that `boring-sys` is able to compile BoringSSL and that rustc can compile Rust code that should be using the library.

.github/workflows/ci.yml
ilammy: Hm... I think `check` would work fine here too, as there is no linker step. Rust does not really support producing cdylibs for iOS targets, only staticlibs, and there is no linkage there. `boring` crate itself does not provide `staticlib`, so this only builds an `rlib`, which is not really useful for building iOS targets.

boring/src/pkey.rs
ilammy: I'm sorry for mixing changes. Moved this commit into #61.

.github/workflows/ci.yml
ilammy: You know what, I have a better idea. How about building the tests, just `--no-run` them? Doing `cargo check` will omit all `#[cfg(test)]` code from compilation – which should be still checked by Clippy, I think, but nevertheless. Maybe some time later someone thinks up a way how to run the binaries found in the target directory.

.github/workflows/ci.yml
ilammy: Also, this reminds me. When CI does `cargo test`, it compiles for the host. That is, I believe that tests for Android targets are actually testing on the host Linux – not the Android 🤔

ilammy:
I've pushed one more commit to work around https://github.com/rust-lang/rust-bindgen/issues/1651 on CI.

.github/workflows/ci.yml
jyn514: If you want to check tests too you can use `cargo check --all-targets`.

ilammy:
I've rebased the branch to resolve some conflicts with recently merged FIPS support.

.github/workflows/ci.yml
ilammy: `cargo test --no-run` with `cargo check --all-targets` then.

jyn514:
@cjmakes do you have a Mac? Can you test out https://github.com/ilammy/boring-example using the instructions in its readme?

eager-signal:
@ilammy @jyn514 I ran through`boring-example` with the iOS Simulator on an M1 Mac and an iOS device, and both worked just fine 👍 .

avently:
@ilammy, could you do the same magic for arm and aarch64 (Linux) target? Can't cross-compile for these platforms from x86_64 Linux host. Looks like it's not supported since https://github.com/cloudflare/quiche/blob/master/quiche/src/build.rs has specific parts related to arm/arm64 Linux while boring-sys doesn't have such parts.

I can test on aarch64 if you don't have such device.

jyn514 (approved):
Sorry for the long delay. Could have sworn I merged this ...

PR #61, opened 2021-12-31
Suppress Clippy warnings about missing safety docs

ilammy:
I'm not quite sure why these are unsafe traits, probably to prevent implementing them for random types accidentally. However, Clippy demands a `# Safety` section in their docs. Tell it to get lost.


jyn514 (approved):
This seems fine for now - I'd like to investigate at some point why these are public, but we should definitely get CI fixed in the meantime. Thanks for the PR :)

Issue #62, opened 2022-01-01
CI should test cross-compilation correctly


ilammy:
Right now GitHub Actions do not do what they should do. Instead of testing for specified targets, they build tests for the host OS and run them there. When proper targets are used, CI fails to compile the tests for various reasons.

- Android targets need to get Rust code linked with Android linker from Android NDK, not system linker.
- Non-native Linux targets need gcc/clang cross-compilers installed for BoringSSL to compile.
- MinGW target for Windows needs either a cross-compiler on Linux, or some more work on Windows.

PR #63, opened 2022-01-01
Explicit run mode and `--target` for CI tests

ilammy:
This is a demonstration of issue #62.

I'm not really sure what do you want to do with these tests. Right now they are broken and it's not trivial to fix them. Might as well leave them broken as a reminder, or remove them altogether from `ci.yml` to be re-added when they actually work.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,54 +82,70 @@ jobs:
           target: x86_64-unknown-linux-gnu
           rust: stable
           os: ubuntu-latest
+          test-type: run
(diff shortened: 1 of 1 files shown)


jyn514:
These were added in https://github.com/cloudflare/boring/commit/ec45baa6d85cfb964b5c0db9a94aad2b5a48259c. @inikulin I think I probably lean towards removing them for now - do you have a strong opinion? We still have tests on the host platform, and it seems pretty difficult to try and get this working when cross-compiling ...

PR #64, opened 2022-01-31
Don't return shutdown errors during shutdown

nox:


tokio-boring/src/lib.rs
```diff
@@ -254,6 +254,19 @@ where
...
             }
             Err(e) => {
+                // If boring returns PROTOCOL_IS_SHUTDOWN then the connection
+                // has already been shutdown and we can just return Ok(()), as
+                // this was exactly what we wanted to do anyway.
+                if e.code() == ErrorCode::SSL {
+                    if let Some(stack) = e.ssl_error() {
+                        if let Some(first) = stack.errors.first() {
+                            if first.code() as i32 == boring_sys::SSL_R_PROTOCOL_IS_SHUTDOWN {
+                                return Poll::Ready(Ok(()));
```
jyn514, line 264: Hmm, could we do this in `boring::ssl::SslStream` instead of `tokio-boring`? It seems like something that's generally useful, not just when using tokio.

jyn514 (commented):
Is this something we can test somehow? When did you run into this error?

tokio-boring/src/lib.rs
nox: `boring::ssl::SslStream` felt more low-level to me, are you sure you want me to move it there?

nox:
> When did you run into this error?

In production with real-world websites.

tokio-boring/src/lib.rs
nox: Done.

boring/src/ssl/mod.rs
```diff
@@ -3054,7 +3054,24 @@ impl<S: Read + Write> SslStream<S> {
...
+                let err = self.make_error(n);
+
+                // If boring returns PROTOCOL_IS_SHUTDOWN then the connection
+                // has already been shutdown and we can just return Ok(()), as
+                // this was exactly what we wanted to do anyway.
+                if e.code() == ErrorCode::SSL {
+                    if let Some(stack) = e.ssl_error() {
+                        if let Some(first) = stack.errors.first() {
+                            if first.code() as i32 == boring_sys::SSL_R_PROTOCOL_IS_SHUTDOWN {
+                                return Poll::Ready(Ok(ShutdownResult::Received));
```
nox, line 3067: Not exactly sure what this means. This is why I didn't put it in the boring crate at first.

kornelski:
I've rebased it and added a safe API for ERR_peek_last_error

PR #67, opened 2022-02-12
boring: `BigNumRef::to_vec_padded()`

ilammy:
Wrap `BN_bn2bin_padded()` which comes useful for exporting fixed-length BIGNUMs, more efficient than padding result of `to_vec()` afterwards.

Note that in OpenSSL the function is called `BN_bn2binpad()` and has a different order of arguments. BoringSSL's BN_bn2bin_padded() also takes the desired length as `size_t`.

### DIFF
--- a/boring/src/bn.rs
+++ b/boring/src/bn.rs
@@ -852,6 +852,36 @@ impl BigNumRef {
         v
     }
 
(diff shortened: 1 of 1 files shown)


jyn514 (approved):
Looks good, thanks for the PR :)

PR #68, opened 2022-02-19
Expose the EVP_AEAD Interface

### LINKED ISSUE #58: Expose the EVP_AEAD Interface
Right now, the [`EVP_AEAD` Interface](https://commondatastorage.googleapis.com/chromium-boringssl-docs/aead.h.html) is the only way to access modern primitives like AES-GCM-SIV and XChaCha20-Poly1305. Exposing it would make it easier for folks to use these primitives in their rust programs.

For use in a different project, I wrote a [quick patch](https://github.com/sgmenda/boring/commit/49d046b0892b95845a1d15de506fd487fe368d46) that exposes `EVP_AEAD` a way heavily inspired by how [`symm.rs`](https://github.com/cloudflare/boring/blob/c4e1f474973744b32579797c1ff51335002a4491/boring/src/symm.rs) currently exposes `EVP_CIPHER`.

If you are interested in this, I can clean up my patch (write more examples, come up with a better name than `AeadCrypter`,...) and submit a PR.

sgmenda:
Fix #58

### DIFF
--- a/boring/src/aead.rs
+++ b/boring/src/aead.rs
@@ -110,8 +110,6 @@ unsafe impl Sync for Aead {}
 unsafe impl Send for Aead {}
 
 /// Represents an AEAD context.
-///
-/// TODO: pick a better name!!
 pub struct AeadCrypter {
     ctx: *mut ffi::EVP_AEAD_CTX,
     max_overhead: usize,
@@ -281,7 +279,61 @@ mod tests {
     }
 
     #[test]
-    fn test_aes_128_gcm_nist_test_vector() {
+    fn test_xchacha20_poly1305_rfc_test_vector() {
+        // Test Vector from https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-xchacha#appendix-A.1
+        let key = Vec::from_hex("808182838485868788898a8b8c8d8e8f909192939495969798999a9b9c9d9e9f")
+            .unwrap();
+        let nonce = Vec::from_hex("404142434445464748494a4b4c4d4e4f5051525354555657").unwrap();
+        let aad = Vec::from_hex("50515253c0c1c2c3c4c5c6c7").unwrap();
+        let data = Vec::from_hex("4c616469657320616e642047656e746c656d656e206f662074686520636c617373206f66202739393a204966204920636f756c64206f6666657220796f75206f6e6c79206f6e652074697020666f7220746865206675747572652c2073756e73637265656e20776f756c642062652069742e").unwrap();
+
+        let expected_tag = Vec::from_hex("c0875924c1c7987947deafd8780acf49").unwrap();
+        let expected_encrypted_data = Vec::from_hex("bd6d179d3e83d43b9576579493c0e939572a1700252bfaccbed2902c21396cbb731c7f1b0b4aa6440bf3a82f4eda7e39ae64c6708c54c216cb96b72e1213b4522f8c9ba40db5d945b11b69b982c1bb9e3f3fac2bc369488f76b2383565d3fff921f9664c97637da9768812f615c68b13b52e").unwrap();
+
+        let encrypted_data =
+            encrypt_aead(Aead::xchacha20_poly1305(), &key, &nonce, &aad, &data).unwrap();
+        assert_eq!(
+            encrypted_data,
+            [expected_encrypted_data, expected_tag].concat()
+        );
+
+        let decrypted_data = decrypt_aead(
+            Aead::xchacha20_poly1305(),
+            &key,
+            &nonce,
+            &aad,
+            &encrypted_data,
+        )
+        .unwrap();
+        assert_eq!(decrypted_data, data);
+    }
+
+    #[test]
+    fn test_aes_128_gcm_siv_rfc_test_vector() {
+        // Test Vector from https://datatracker.ietf.org/doc/html/rfc8452#appendix-C.1
+        let key = Vec::from_hex("f901cfe8a69615a93fdf7a98cad48179").unwrap();
+        let nonce = Vec::from_hex("6245709fb18853f68d833640").unwrap();
+        let aad =
+            Vec::from_hex("7576f7028ec6eb5ea7e298342a94d4b202b370ef9768ec6561c4fe6b7e7296fa859c21")
+                .unwrap();
+        let data = Vec::from_hex("e42a3c02c25b64869e146d7b233987bddfc240871d").unwrap();
+
+        let expected_encrypted_data = Vec::from_hex(
+            "391cc328d484a4f46406181bcd62efd9b3ee197d052d15506c84a9edd65e13e9d24a2a6e70",
+        )
+        .unwrap();
+
+        let encrypted_data =
+            encrypt_aead(Aead::aes_128_gcm_siv(), &key, &nonce, &aad, &data).unwrap();
+        assert_eq!(encrypted_data, expected_encrypted_data);
+
+        let decrypted_data =
+            decrypt_aead(Aead::aes_128_gcm_siv(), &key, &nonce, &aad, &encrypted_data).unwrap();
+        assert_eq!(decrypted_data, data);
+    }
+
+    #[test]
+    fn test_aes_256_gcm_nist_test_vector() {
         // A NIST AES-GCM Test Vector from https://csrc.nist.gov/projects/cryptographic-algorithm-validation-program/CAVP-TESTING-BLOCK-CIPHER-MODES#GCMVS
 
         let key = Vec::from_hex("78e53ed07c0f162406ee17c54344e2ae").unwrap();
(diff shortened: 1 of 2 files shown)


boring/src/aead.rs
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    /// Returns the length of keys used with this AEAD.
+    pub fn key_len(&self) -> usize {
+        unsafe { EVP_AEAD_key_length(self.0) as usize }
+    }
+
+    /// Returns the length of nonces used with this AEAD.
+    pub fn nonce_len(&self) -> usize {
+        unsafe { EVP_AEAD_nonce_length(self.0) as usize }
```
jyn514, line 91: Casting here doesn't seem correct - can you change this to return `libc::size_t` instead? Same for all the other functions returning `usize`.

boring/src/aead.rs
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    /// Returns the maximum ciphertext overhead with this AEAD.
```
jyn514, line 94: Let's use the upstream documentation here, I'd prefer not to reuse the word "overhead" when defining what it means.
```suggestion
    /// Returns the maximum number of additional bytes added by the act of sealing data with `self`.
    ///
    /// Corresponds to [`EVP_AEAD_max_overhead`](https://commondatastorage.googleapis.com/chromium-boringssl-docs/aead.h.html#EVP_AEAD_max_overhead).
```

boring/src/aead.rs
```diff
@@ -0,0 +1,357 @@
...
+        unsafe { EVP_AEAD_max_overhead(self.0) as usize }
+    }
+
+    /// Returns the maximum tag length with this AEAD.
+    pub fn max_tag_length(&self) -> usize {
+        unsafe { EVP_AEAD_max_tag_len(self.0) as usize }
+    }
+
+    pub fn as_ptr(&self) -> *const ffi::EVP_AEAD {
+        self.0
```
jyn514, line 105: boring does this elsewhere in the codebase with `ForeignType` - is there a reason you didn't use that here? https://docs.rs/boring/latest/boring/dh/struct.Dh.html#impl-ForeignType

boring/src/aead.rs
```diff
@@ -0,0 +1,357 @@
...
+    }
+}
+
+unsafe impl Sync for Aead {}
+unsafe impl Send for Aead {}
```
jyn514, line 110: How do you know the cipher is thread-safe?

boring/src/aead.rs
```diff
@@ -0,0 +1,357 @@
...
+
+    /// Seals data by encrypting and authenticating the provided data and authenticating the provided aad.
+    ///
+    /// Returns the number of bytes written to output.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `output.len() > c_int::max_value()`.
+    ///
+    /// Panics if `output.len() < data.len() + aead.max_overhead()`.
```
jyn514, line 146: I know this mimics the interface of the underlying library, but can we provide a better one here? I think it's fine to allocate a new buffer instead of taking an `output` argument - that matches the behavior of [`DH::params_to_pem`](https://docs.rs/boring/latest/boring/dh/struct.Dh.html#method.params_to_pem).

boring/src/aead.rs
```diff
@@ -0,0 +1,357 @@
...
+    /// Panics if `output.len() > c_int::max_value()`.
+    ///
+    /// Panics if `output.len() < data.len()`.
+    pub fn open(
+        &mut self,
+        nonce: &[u8],
+        aad: &[u8],
+        data: &[u8],
+        output: &mut [u8],
+    ) -> Result<usize, ErrorStack> {
```
jyn514, line 195: Same here, it would be better to return an owned buffer.

boring/src/aead.rs
```diff
@@ -0,0 +1,357 @@
...
+    pub fn as_ptr(&self) -> *const ffi::EVP_AEAD {
+        self.0
+    }
+}
+
+unsafe impl Sync for Aead {}
+unsafe impl Send for Aead {}
+
+/// Represents an AEAD context.
+pub struct AeadCrypter {
```
jyn514, line 113: you mentioned wanting to come up with a better name - I think `AeadContext` would be a good fit. But this also seems fine as-is.

boring/src/aead.rs
```diff
@@ -0,0 +1,357 @@
...
+        unsafe {
+            ffi::EVP_AEAD_CTX_free(self.ctx);
+        }
+    }
+}
+
+/// Encrypts the provided data and authenticates the ciphertext and the provided aad.
+///
+/// Returns the sealed data as a vector.
+pub fn encrypt_aead(
```
jyn514, line 234: should this be an inherent method on `Aead` maybe?

boring/src/aead.rs
```diff
@@ -0,0 +1,357 @@
...
+    aead: Aead,
+    key: &[u8],
+    nonce: &[u8],
+    aad: &[u8],
+    data: &[u8],
+) -> Result<Vec<u8>, ErrorStack> {
+    let mut aeadcrypter = AeadCrypter::new(aead, key)?;
+    let mut output = vec![0u8; data.len() + aead.max_ciphertext_overhead()];
+    let bytes_written_to_output = aeadcrypter.seal(nonce, aad, data, &mut output)?;
+    Ok(output[..bytes_written_to_output].to_vec())
```
jyn514, line 244:
```suggestion
    output.truncate(bytes_written_to_output);
    Ok(output)
```

boring/src/aead.rs
```diff
@@ -0,0 +1,357 @@
...
+}
+
+/// Authenticates the provided data and aad and decrypts the provided data.
+///
+/// Returns the unsealed data as a vector.
+pub fn decrypt_aead(
```
jyn514, line 250: ditto, this seems better as an inherent method.

boring/src/aead.rs
```diff
@@ -0,0 +1,357 @@
...
+mod tests {
+    use super::*;
+    use hex::{self, FromHex};
+
+    #[test]
+    fn test_aead_encrypt_decrypt() {
+        let key = [0u8; 16];
+        let nonce = [0u8; 16];
+        let aad = [0u8; 16];
+        let data = [0u8; 16];
```
jyn514, line 273: Could you use something other than zeros here? I'm not sure what boringssl does on zeros, and it seems better to be closer to a real scenario anyway.

jyn514 (changes requested):
This looks like a good start :) I left a lot of comments - none of them are critical, but it would be nice to address at least some before merging.

Could you add `EVP_aead_aes_256_gcm_randnonce` and `EVP_aead_aes_128_gcm_randnonce` under a `cfg(feature = "fips")` flag? `EVP_aead_aes_128_ccm_bluetooth`, `EVP_aead_aes_128_ccm_bluetooth_8`, and `EVP_has_aes_hardware` also seem to be missing, I think you can add those unconditionally.

boring/src/aead.rs
```diff
@@ -0,0 +1,357 @@
...
+//!     aead,
+//!     key,
+//!     nonce,
+//!     aad,
+//!     data).unwrap();
+//!
+//! assert_eq!(
+//!     b"\x3D\xCC\xF8\x9A\x2B\xC1\x91\x15\xCD\x19\xD2\xC2\xB1\x13\x01\x38\xC5\
+//!       \x7D\xF5\xEE\x5B\x37\xD1\x44\x05\xD2\x1A\xDA\x3F\x69\x97\x5D",
+//!     &sealed_data[..]);
```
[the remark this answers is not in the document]
sgmenda: `sealed_data` is a `Vec<u8>` while `b"[...]"` is a [u8; 32] so we need the `[..]`.

boring/src/aead.rs
sgmenda: I tried to maintain parity with `symm.rs` which uses this pattern when returning lengths: https://github.com/cloudflare/boring/blob/e6ddc403337de031e446607cd161bcc41f2610db/boring/src/symm.rs#L187-L191

boring/src/aead.rs
sgmenda: I agree. I defined them separately for parity with `symm.rs`.

https://github.com/cloudflare/boring/blob/e6ddc403337de031e446607cd161bcc41f2610db/boring/src/symm.rs#L635-L685

boring/src/aead.rs
sgmenda: Good catch, I don't I accidentally assumed that it was, and the [the documentation](https://commondatastorage.googleapis.com/chromium-boringssl-docs/aead.h.html) doesn't seem explicit on this, so I'll remove the lines.

boring/src/aead.rs
sgmenda: I tried to follow the one in `symm.rs`.

https://github.com/cloudflare/boring/blob/e6ddc403337de031e446607cd161bcc41f2610db/boring/src/symm.rs#L182-L185

boring/src/aead.rs
sgmenda: I agree, I was trying to match `symm.rs`.

https://github.com/cloudflare/boring/blob/e6ddc403337de031e446607cd161bcc41f2610db/boring/src/symm.rs#L446-L482

sgmenda:
Thank you for detailed comments! I made a few of the suggested changes; the others seem to break parity with `symm.rs` which afaict seems okay, but I wanted to confirm with you before making the change.

PR #70, opened 2022-04-27
Update boringssl deps for quiche 0.12.0 compatibility

BiagioFesta:
Latest release of `quiche` (`0.12.0`) [uses boringssl commit: `f1c7534`](https://github.com/cloudflare/quiche/tree/0.12.0/quiche/deps).

This breaks case where one wants to make `quiche` and `boring` rely on the same `quichessl` implementation (via setting `BORING_BSSL_PATH` env var).

```
          ┌───────────┐
    ┌─────┤ User Code ├──────┐
    │     └───────────┘      │
    │                        │
    ▼                   ┌────▼────┐
┌────────┐              │ boring  │ 
│ Quiche │              └────┬────┘
└───┬────┘                   │
    │                   ┌────▼─────┐
    │                   │boring-sys│
    │                   └────┬─────┘
    │     ┌───────────┐      │
    └─────► boringssl │◄─────┘
          └───────────┘
```

That's because of some changes in the public interface of `boringssl`.

In particular,

- ExtensionType::TOKEN_BINDING removed from implementation
https://github.com/google/boringssl/commit/d89ec688f2296f41b4999b9db6e8bfba79c18050
> We didn't end up deploying this. We also never implemented the final
RFC, so what we do have isn't useful for someone who wishes to deploy
it anyway.

- ExtensionType::ECH_IS_INNER removed from implementation
https://github.com/google/boringssl/commit/18b6836b2f6340187a7981c82be9be9d092d36d6
> [...]
> ech_is_inner is now folded into the main encrypted_client_hello code
  point. This works better with some stuff around HRR.

---

This CR is going to update `boringssl` dependency and use the same commit from latest release of `quiche`, adapting the interface changes in the `boring` crate.

### DIFF
--- a/boring/src/ssl/mod.rs
+++ b/boring/src/ssl/mod.rs
@@ -486,7 +486,6 @@ impl ExtensionType {
         Self(ffi::TLSEXT_TYPE_application_layer_protocol_negotiation as u16);
     pub const PADDING: Self = Self(ffi::TLSEXT_TYPE_padding as u16);
     pub const EXTENDED_MASTER_SECRET: Self = Self(ffi::TLSEXT_TYPE_extended_master_secret as u16);
-    pub const TOKEN_BINDING: Self = Self(ffi::TLSEXT_TYPE_token_binding as u16);
     #[cfg(not(feature = "fips"))]
     pub const QUIC_TRANSPORT_PARAMETERS_LEGACY: Self =
         Self(ffi::TLSEXT_TYPE_quic_transport_parameters_legacy as u16);
@@ -511,8 +510,6 @@ impl ExtensionType {
     pub const APPLICATION_SETTINGS: Self = Self(ffi::TLSEXT_TYPE_application_settings as u16);
     #[cfg(not(feature = "fips"))]
     pub const ENCRYPTED_CLIENT_HELLO: Self = Self(ffi::TLSEXT_TYPE_encrypted_client_hello as u16);
-    #[cfg(not(feature = "fips"))]
-    pub const ECH_IS_INNER: Self = Self(ffi::TLSEXT_TYPE_ech_is_inner as u16);
     pub const CERTIFICATE_TIMESTAMP: Self = Self(ffi::TLSEXT_TYPE_certificate_timestamp as u16);
     pub const NEXT_PROTO_NEG: Self = Self(ffi::TLSEXT_TYPE_next_proto_neg as u16);
     pub const CHANNEL_ID: Self = Self(ffi::TLSEXT_TYPE_channel_id as u16);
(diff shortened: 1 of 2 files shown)


BiagioFesta:
Sure.

I have been digging into the code (google/boringssl) and the failure of the test is basically caused by: https://boringssl.googlesource.com/boringssl/+/c02c19e0d842f54d903a9b62316476f4b9c4e3f0

---

The boring (rust crate)'s test relies on the "old" behavior: where returning `SSL_TLSEXT_ERR_ALERT_FATAL` from callback of ALPN server selection produces no error in SSL API and `selected_alpn_protocol()` [is just an empty `Option`](https://github.com/cloudflare/boring/blob/e6ddc403337de031e446607cd161bcc41f2610db/boring/src/ssl/test/mod.rs#L457).

Now the behavior in boringssl has been changed (fixed) in order to be *RFC7301* compliant.

When no intersection happens betweens client and server ALPNs (and client set is not empty),  server replies with an alert message with code 120 (No application protocol found).
That's when the callback for ALPN return (like in this test): `SSL_TLSEXT_ERR_ALERT_FATAL`. 

An error is also now returned in the SSL API in that case.

---

Therefore the test is wrong. 

When ALPN server validation returns `SSL_TLSEXT_ERR_ALERT_FATAL`, we should indeed expect an error in the TLS connection on both server and client error (the server terminate the connection with an alert).

I have also confirmed by looking at message exchange in both situations.
image


Please also note that in case the validation is "not fatal" and server wants to continues with the SERVER HELLO message (with an empty ALPN intersection), a different returning value is used, i.e., `SSL_TLSEXT_ERR_NOACK`.

Indeed, a [test below the failing one](https://github.com/cloudflare/boring/blob/e6ddc403337de031e446607cd161bcc41f2610db/boring/src/ssl/test/mod.rs#L460-L472) is perfectly fine.


I am going to try to fix the test and update my PR.

jyn514:
> The boring (rust crate)'s test relies on the "old" behavior: where returning SSL_TLSEXT_ERR_ALERT_FATAL from callback of ALPN server selection produces no error in SSL API and selected_alpn_protocol() [is just an empty Option](https://github.com/cloudflare/boring/blob/e6ddc403337de031e446607cd161bcc41f2610db/boring/src/ssl/test/mod.rs#L457).

> Now the behavior in boringssl has been changed (fixed) in order to be RFC7301 compliant.

@LPardue @gabi-250 @inikulin do we have any code that depends on this behavior?

BiagioFesta:
Thanks for merging! any chance a new release could go on [crates.io](http://crates.io/) so that we can get both this and quiche from it?

jyn514:
I will try to make a release in the next few weeks; I have other more pressing obligations at the moment.

PR #72, opened 2022-06-02
boring-sys: Save time by not building the bssl CLI utility

jrose-signal:
We only need libcrypto and libssl; nothing in the project uses `bssl`, and it's not exposed to downstream clients.

### DIFF
--- a/boring-sys/build.rs
+++ b/boring-sys/build.rs
@@ -263,7 +263,8 @@ fn main() {
             cfg.define("FIPS", "1");
         }
 
-        cfg.build_target("bssl").build().display().to_string()
+        cfg.build_target("ssl").build();
+        cfg.build_target("crypto").build().display().to_string()
     });
 
     let build_path = get_boringssl_platform_output_path();


boring-sys/build.rs
```diff
@@ -263,7 +263,8 @@ fn main() {
             cfg.define("FIPS", "1");
         }
 
-        cfg.build_target("bssl").build().display().to_string()
+        cfg.build_target("ssl").build();
+        cfg.build_target("crypto").build().display().to_string()
```
jrose-signal, line 267: Using CMake list syntax `"ssl;crypto"` also works, and in theory allows for more parallelism, but since it's not a documented feature of the `cmake` crate I went with two build invocations instead.

jyn514 (approved):
Thanks!

boring-sys/build.rs
jyn514: Does using `build_target("ssl").build_target("crypto").build ()` regain the parallelism?

jrose-signal: Afraid not; it "sets" the build target rather than "adding" one ([confirmed by looking at the source](https://docs.rs/cmake/latest/src/cmake/lib.rs.html#367-370)).

jrose-signal: Testing `cargo +nightly build --release --timings -p boring-sys` and looking at just the time spent running build.rs:

| master | This PR | `"ssl;crypto"` |
|:-:|:-:|:-:|
| 56s | 49s | 47s |

I think that last 2s is probably within experimental variation. (Testing on a 4-core Intel MacBook Pro.)

PR #73, opened 2022-06-03
Add support for Mac Catalyst (on Intel and Apple Silicon chips)

jrose-signal:
Builds on #60 to support compilation for Mac Catalyst targets as well (iOS-on-macOS). These are Tier 3 targets, but apart from working around a hardcoded version in the 'cc' crate everything does work.

Recommended to review with whitespace diffing disabled.

### DIFF
--- a/boring-sys/build.rs
+++ b/boring-sys/build.rs
@@ -70,6 +70,20 @@ const CMAKE_PARAMS_IOS: &[(&str, &[(&str, &str)])] = &[
             ("CMAKE_OSX_SYSROOT", "iphonesimulator"),
         ],
     ),
+    (
+        "aarch64-apple-ios-macabi",
+        &[
+            ("CMAKE_OSX_ARCHITECTURES", "arm64"),
+            ("CMAKE_OSX_SYSROOT", "macosx"),
+        ],
+    ),
+    (
+        "x86_64-apple-ios-macabi",
+        &[
+            ("CMAKE_OSX_ARCHITECTURES", "x86_64"),
+            ("CMAKE_OSX_SYSROOT", "macosx"),
+        ],
+    ),
 ];
 
 fn cmake_params_ios() -> &'static [(&'static str, &'static str)] {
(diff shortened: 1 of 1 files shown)


boring-sys/build.rs
```diff
@@ -178,17 +193,29 @@ fn get_boringssl_cmake_config() -> cmake::Config {
...
-            let target_cflag = if arch == "x86_64" {
-                "-target x86_64-apple-ios-simulator"
+            if target.ends_with("-macabi") {
+                // Mac Catalyst
+                let compiler_flags = format!("{} -target {}", bitcode_cflag, target);
+                boringssl_cmake.define("CMAKE_ASM_FLAGS", &compiler_flags);
+                // Work around hardcoded deployment target in cc crate by defining CMAKE_C_FLAGS
+                // instead of using the cflag builder.
+                boringssl_cmake.define("CMAKE_C_FLAGS", &compiler_flags);
+                boringssl_cmake.define("CMAKE_CXX_FLAGS", &compiler_flags);
```
jrose-signal, line 203: rust-lang/cc-rs#471 hardcoded the `*-apple-ios-macabi` targets to use a minimum OS version of "13.0", but the Clang provided in the latest Xcode only supports a minimum OS version of "13.1". rust-lang/cc-rs#678 discusses a possible fix for this, but it's not a clear win.

But we don't need to wait for them; by changing to `-target x86_64-apple-ios-macabi` (or `aarch64-apple-ios-macabi`), we let the compiler pick the earliest possible version, which is sufficient for compiling boring-ssl even if it isn't sufficient for all Rust use cases.

This isn't ideal because it means other `cflags` won't work, and it's not a *complete* solution because `cargo test` doesn't work (linking an executable still goes through the cc crate). But manually linking and running the tests shows them passing, and our experiments with boring-sys in [signalapp/libsignal](/signalapp/libsignal) show the library working correctly when linked statically.

inikulin: May I ask you to add a user-friendly remark about that in crate's docs in `boring/lib.rs`?

jrose-signal: Catalyst stuff seems to have gotten better in Rust 1.71, which is slightly ahead of the toolchain we're currently using, so I'll hang on to this a bit longer and maybe the workaround will become unnecessary.

jrose-signal:
Okay, between the improvements on the Rust side and Signal itself deciding not to pursue Catalyst at this time, I think we can close this.

PR #76, opened 2022-06-08
Use Ninja to build BoringSSL

msoxzw:
[Ninja build system is recommended by BoringSSL](https://github.com/google/boringssl/blob/master/BUILDING.md). Furthermore, the MSVC generator hack is no longer required, and the MinGW-w64 toolchain on Windows also works, the performance of which is presumably better and even on par with Linux, according to [Agner](https://www.agner.org/optimize/blog/read.php?i=1015) and [simdjson](https://github.com/simdjson/simdjson/blob/master/doc/performance.md).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -151,6 +151,8 @@ jobs:
       run: rustup update ${{ matrix.rust }} --no-self-update && rustup default ${{ matrix.rust }}
       shell: bash
     - run: rustup target add ${{ matrix.target }}
+    - name: Install Ninja
+      uses: seanmiddleditch/gha-setup-ninja@v3
     - name: Install nasm
       if: startsWith(matrix.os, 'windows')
       run: choco install nasm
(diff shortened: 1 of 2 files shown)


jyn514:
Won't this require users to have ninja installed? I'm reluctant to merge this if it's going to break people's CI.

msoxzw:
Yes, Ninja is required to install, and I am attempting to fix CI failure.

olix0r:
@msoxzw If this is merged, it will break the builds of all crates that depend on boring/tokio-boring unless their build environments already have ninja installed. So, at the very least, this should probably have to be clearly marked as a breaking change.

Is it possible to retain the current build behavior and support an alternate build mode that uses Ninja, i.e., when an environment variable is set?

msoxzw:
@olix0r Definitely, [CMAKE_GENERATOR environment variable](https://cmake.org/cmake/help/latest/envvar/CMAKE_GENERATOR.html) could be used for this purpose, but the MSVC generator hack have to be reserved, and I also wonder how to modify CI to determine which build system should be used, or both.

jyn514:
> [CMAKE_GENERATOR environment variable](https://cmake.org/cmake/help/latest/envvar/CMAKE_GENERATOR.html) could be used for this purpose

Does that mean people can already opt-in to ninja today using that variable? In that case I think we should close this, rather than overriding the generator that people pick.

PR #80, opened 2022-07-12
Put BoringSSL's SSL/TLS support behind a new feature 'ssl'

jrose-signal:
BoringSSL vends two libraries: libcrypto and libssl. Some clients don't need the libssl part. With a feature flag, we can skip building it (saving on compile time) and initializing it (saving on code size and a tiny bit of run time).

The new feature is on by default for backwards compatibility.

The one thing I wasn't sure what to do with was CI. Should it be a single alternate configuration, like FIPS, or something more? Do we care about FIPS-without-libssl?

### DIFF
--- a/boring-sys/Cargo.toml
+++ b/boring-sys/Cargo.toml
@@ -31,5 +31,8 @@ bindgen = { version = "0.60", default-features = false, features = ["runtime"] }
 cmake = "0.1"
 
 [features]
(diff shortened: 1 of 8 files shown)


jrose-signal:
We're less interested in this now that we're also using libssl, so I'm going to close this unless someone else wants it.

PR #81, opened 2022-07-20
boring: Support validating certificates against CRLs

ravi-signal:
Adds CRL types and validation of certificates against CRLs with
a `X509Store/X509StoreContext`. Also enables customizing verification
flags on the `X509Store`, required to enable CRL checking.

Apologies for dropping such a large PR here without discussion first — please let me know if you'd prefer an alternate approach, I'm happy to rework this.


jrose-signal:
Rebased on master. We have some follow-up work after this goes in to improve support for certificate extensions, but it depends what you want to do with this first.

PR #82, opened 2022-07-25
boring-sys: Use the Android NDK sysroot when running bindgen

jrose-signal:
Things kinda work without it on Linux if you have `gcc-multilib` installed, but that's just getting lucky. We compile against the Android sysroot; we should use it for bindgen as well.

### DIFF
--- a/boring-sys/build.rs
+++ b/boring-sys/build.rs
@@ -288,6 +288,16 @@ fn get_extra_clang_args_for_bindgen() -> Vec<String> {
             params.push("-isysroot".to_string());
             params.push(sysroot);
         }
+        "android" => {
+            let android_ndk_home = std::env::var("ANDROID_NDK_HOME")
+                .expect("Please set ANDROID_NDK_HOME for Android build");
+            let mut android_sysroot = std::path::PathBuf::from(android_ndk_home);
+            android_sysroot.push("sysroot");
+            params.push("--sysroot".to_string());
+            // If ANDROID_NDK_HOME weren't a valid UTF-8 string,
+            // we'd already know from std::env::var.
+            params.push(android_sysroot.into_os_string().into_string().unwrap());
+        }
         _ => {}
     }
 


boring-sys/build.rs
```diff
@@ -288,6 +288,14 @@ fn get_extra_clang_args_for_bindgen() -> Vec<String> {
             params.push("-isysroot".to_string());
             params.push(sysroot);
         }
+        "android" => {
+            let android_ndk_home = std::env::var("ANDROID_NDK_HOME")
+                .expect("Please set ANDROID_NDK_HOME for Android build");
```
jyn514, line 293: This is now required when setting `BORING_BSSL_PATH` to use a pre-compiled boringssl object file. Is that intentional? It seems reasonable for these requirements to match, but I don't use Android so I don't know how necessary the NDK is for compiling code.

boring-sys/build.rs
```diff
@@ -288,6 +288,14 @@ fn get_extra_clang_args_for_bindgen() -> Vec<String> {
...
+            let mut android_sysroot = std::path::PathBuf::from(android_ndk_home);
+            android_sysroot.push("sysroot");
+            params.push("--sysroot".to_string());
+            params.push(android_sysroot.to_string_lossy().into());
```
jyn514, line 297: to_string_lossy looks wrong. This should either be an `unwrap` (I doubt anyone is using non-unicode paths in practice) or we should pass byte strings to the bindgen invocation.

jyn514 (changes requested):
Seems reasonable. Have you tested these changes? I don't have a way of testing android.

boring-sys/build.rs
jrose-signal: That's an interesting point. Another way to phrase this is "how much do the system C headers affect BoringSSL's public headers" and I haven't looked into that at all. C is a finicky language, but I'd also expect BoringSSL's public API surface to be pretty, well, boring.

boring-sys/build.rs
jrose-signal: Bindgen doesn't accept byte strings. :sob: I'll make it an `unwrap`, or rather an `expect`; you're right that substituting U+FFFD is never going to make a valid path and only obscures the issue.

boring-sys/build.rs
jrose-signal: I will say that the further the host platform is from the Android platform being targeted, though, the more likely there'll be a problem.

jyn514: ah I see, the way this can go wrong is something like
```c
 int ASN1_TYPE_get(const ASN1_TYPE *a);
```
and `int` has a different size depending on whether you're using the Android SDK or not. Ok, I agree we should require this even if it leads to build failures in rare cases; better that than to silently use the wrong ABI.

jrose-signal:
Testing: have tested that this builds successfully, but have not run on actual Android devices yet.

boring-sys/build.rs
jrose-signal: Even better, `std::env::var` already checked for us.

boring-sys/build.rs
jrose-signal: Right. Usually even that will be fine because bindgen will generate `libc::c_int`, but that's the type of thing that could get messed up.

jyn514: makes sense - I worry most about it in macro expressions, bindgen's handling of `#define x y` is ... pretty hacky.

jyn514:
Thanks for the fix!

PR #84, opened 2022-08-02
boring-sys: Always treat macOS targets as a cross-compile, like iOS

jrose-signal:
This fixes actual cross-compile cases, like compiling for Apple Silicon on an Intel Mac. It also makes builds more consistent by using the SDK in the installed Xcode instead of the host root headers, which is how Apple recommends people build anyway. (If you only have the command line tools installed, though, it should fall back to those without a problem.)

In our local branch we have this combined with #73 and #78, but this is pretty separable and probably the least controversial.


inikulin:
@jrose-signal Thank you! Could you rebase please?

inikulin (approved):

PR #85, opened 2022-08-08
hyper-boring: Impl debug for MaybeHttpsStream

4JX:
Formatting is taken straight out of hyper-rustls

### DIFF
--- a/hyper-boring/src/lib.rs
+++ b/hyper-boring/src/lib.rs
@@ -327,7 +327,7 @@ where
     }
 }
 
-impl<T: fmt::Debug> fmt::Debug for MaybeHttpsStream<T> {
+impl<T> fmt::Debug for MaybeHttpsStream<T> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match *self {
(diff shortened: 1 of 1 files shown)


hyper-boring/src/lib.rs
```diff
@@ -326,3 +326,12 @@ where
         }
     }
 }
+
+impl<T: fmt::Debug> fmt::Debug for MaybeHttpsStream<T> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            MaybeHttpsStream::Http(..) => f.pad("Http(..)"),
+            MaybeHttpsStream::Https(..) => f.pad("Https(..)"),
```
jyn514, line 334: Why not show the debug impl for the underlying T as well?

Anyway, this seems good overall - if you either add the underlying impl or remove the `T: Debug` bound I'm happy to merge this.

4JX: Ah that slipped by. I only really needed `MaybeHttpsStream` to implement it.

jyn514:
Thanks!

PR #86, opened 2022-08-22
boring-sys: Add an android-api-19 feature for compat with Android 4.4

jrose-signal:
I tested Signal's use of BoringSSL on an Android 4.4 x86 emulator, though I don't have a device at hand.

Alternate approaches include
- Check an environment variable. More flexible, but also more global state.
- Change the default for everyone. Didn't seem as likely to be upstreamed, but BoringSSL doesn't actually seem to depend on anything between Android 19 and 21 that it can't check dynamically.
- Remove the setting entirely, which would implicitly pick the lowest version for the target (Android 16 for 32-bit). That might have further implications, so I didn't go that route either.


jrose-signal:
(The Clippy issues are pre-existing but should probably also be looked at!)

Noah-Kennedy (approved):

jrose-signal:
This was approved but never merged. We've been using it in our fork, but Signal is now dropping Android 19 too, so we don't have interest in it anymore. I'll leave it to the maintainers to decide whether to merge or close this.

PR #87, opened 2022-09-07
Remove uses of `mem::uninitialized`

jyn514:
According to [the docs](https://doc.rust-lang.org/stable/std/mem/fn.uninitialized.html),

> Calling this when the content is not yet fully initialized causes immediate undefined behavior.
> it [is] undefined behavior to have uninitialized data in a variable even if that variable has an integer type.

Using MaybeUninit instead, as recommended by the official documentation, avoids undefined behavior by not creating a `&mut` reference to uninitialized data.

This also fixes some harmless clippy warnings.

### DIFF
--- a/boring-sys/build.rs
+++ b/boring-sys/build.rs
@@ -42,7 +42,7 @@ fn cmake_params_android() -> &'static [(&'static str, &'static str)] {
     };
     for (android_arch, params) in cmake_params_android {
         if *android_arch == arch {
-            return *params;
+            return params;
         }
     }
     &[]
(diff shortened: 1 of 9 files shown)


inikulin (approved):

jyn514:
The test failure is from the newly added test in https://github.com/cloudflare/boring/pull/70. I see it failing even on the master branch, before this change, but strangely CI for master is passing.

BiagioFesta:
In the commit of this PR: [Fix clippy warnings](https://github.com/cloudflare/boring/pull/87/commits/f8e9860d1cff5a404ffafb5c29207be1acd7e532), I see you are reverting the update of boringssl (sys-dep).

Probably it is a mistake?

The test fails because (briefly speaking) old boringssl expects `Ok` and the new one expects `Err`.
All information are already contained in my CR: https://github.com/cloudflare/boring/pull/70#issuecomment-1112162453

jyn514:
I'm a fool, you're absolutely correct :facepalm: thanks for pointing out my mistake :)

PR #89, opened 2022-09-28
Allow overriding the link search path for `crypto` and `ssl` libraries

AkshatM:
Currently, Boring hardlinks the search folder for `crypto` and `ssl` shared objects as build output subfolders under BORING_BSSL_PATH. However, if BORING_BSSL_PATH is set to point to a pre-built binary, the shared objects don't need to live in the build output folder - and likely won't. 

This commit allows overriding the search folder independently of BORING_BSSL_PATH.

### DIFF
--- a/boring-sys/build.rs
+++ b/boring-sys/build.rs
@@ -363,22 +363,28 @@ fn main() {
         cfg.build_target("crypto").build().display().to_string()
     });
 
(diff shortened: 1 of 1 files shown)


jyn514:
Why not create a `build/` directory yourself?

jyn514:
Going to close this since it's easy to replicate the functionality with existing features.

PR #90, opened 2022-09-29
Implement AsRawFd for SslStream to return -1

quangIO:
Per internal discussion with @inikulin, we should be able to return -1 as the raw fd of SslStream

### DIFF
--- a/boring/src/ssl/mod.rs
+++ b/boring/src/ssl/mod.rs
@@ -71,6 +71,7 @@ use std::io::prelude::*;
 use std::marker::PhantomData;
 use std::mem::{self, ManuallyDrop};
 use std::ops::{Deref, DerefMut};
+use std::os::unix::prelude::{AsRawFd, RawFd};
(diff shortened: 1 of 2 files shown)


quangIO:
> Why is this useful?

The use case is: we have a custom Socket trait. Some sockets are transferable (as in, can be sent to other processes), and some are not. SslStream is a non transferable socket as it manipulates bytestream.

jyn514:
> We are implementing other trait for that, but doing this directly using AsRawFd doesn't hurt.

I don't quite understand; if you're already implementing the other trait, why do you also need AsRawFd? I worry this will be very surprising for people who expect AsRawFd to always return a valid file descriptor.

Noah-Kennedy:
I'm in agreement with @jyn514. `AsRawFd` shouldn't be used for situations like this.

nox:
Closing this as it seems we don't want to do that.

Issue #93, opened 2022-11-05
re-enable decrepit ?


ibigbug:
the commit msg indicates it's removed in: https://github.com/cloudflare/boring/commit/a6f5beeb333342a5e84a3f36ee2f8b80a6f8d602

is it meaningful to bring it back?

PR #94, opened 2022-11-05
add decrepit to link

ibigbug:
#93


jyn514:
Why do you want this change? What does it add that's useful?

ibigbug:
sorry i didnt make it clear - there are certain functions that only exist in decrepit https://github.com/google/boringssl/blob/1510e460c646777346bebb9eb060d4c50e093813/include/openssl/cipher.h#L461

PR #96, opened 2022-11-30
boring-sys: Update Android NDK sysroot path

jrose-signal:
This longer path (inside the prebuilt toolchain included in the NDK) has been the preferred sysroot since NDK r19. Newer NDKs no longer have a top-level "sysroot" directory at all.

Follow-up to #82, my bad for testing against an older NDK at that time.


inikulin:
@jrose-signal sorry for the super-long delay in response. Could you please fix clippy lint failures?

inikulin (approved):

PR #108, opened 2023-03-27
Implement ToOwned for X509StoreRef and Clone for X509Store

nox:


nox:
Apparently clang-7 was removed from Ubuntu 22.04?

ghedo (approved):

PR #118, opened 2023-05-11
fix: BIO_set_retry_write when BIO_CTRL_FLUSH to allow writer returns WouldBlock on flush

ihciah:
A fix for rust-openssl can be found at https://github.com/sfackler/rust-openssl/pull/1922

This PR fixes a BIO_flush related issue. According to openssl doc([link](https://www.openssl.org/docs/man1.0.2/man3/BIO_flush.html)):

BIO_flush(), because it can write data may return 0 or -1 indicating that the call should be retried later in a similar manner to BIO_write(). The BIO_should_retry() call should be used and appropriate action taken is the call fails.

We have to call BIO_set_retry_write on retriable_error like WouldBlock. Otherwise, when the flush returns WouldBlock, it will fail.

Since we usually wrap openssl with a TcpStream, there is no buffer inside and flush always returns Ok(()), this issue is not obvious. However, if users use a buffered io, the flush support will be essential.


kornelski:
Does this require upstream boringssl to fix the status too? https://github.com/openssl/openssl/commit/bcbc7d60679b79fa4347e33c865306dce41ed985

PR #134, opened 2023-07-31
Introduce async callbacks

nox:


boring/src/ssl/connector.rs
```diff
@@ -189,6 +189,10 @@ impl ConnectConfiguration {
         self.verify_hostname = verify_hostname;
     }
 
+    pub fn ssl_mut(&mut self) -> &mut SslRef {
```
nox, line 192: Not sure I like exposing this, but there is no other way to set `TASK_CONTEXT_INDEX` from `tokio_boring::connect` prior to calling `ConnectConfiguration::connect`.

boring/src/ssl/connector.rs
```diff
@@ -324,8 +328,12 @@ impl SslAcceptor {
     where
         S: Read + Write,
     {
-        let ssl = Ssl::new(&self.0)?;
-        ssl.accept(stream)
+        self.new_session()?.accept(stream)
+    }
+
+    /// Creates a new TLS session, ready to accept a stream.
+    pub fn new_session(&self) -> Result<Ssl, ErrorStack> {
```
nox, line 335: There is no way to set `TASK_CONTEXT_INDEX` from `tokio_boring::accept` when calling `SslAcceptor::accept`, and given the `Ssl` value doesn't even exist yet I just can't use `SslAcceptor::accept` anymore in `tokio_boring::accept`.

boring/src/ssl/mod.rs
```diff
@@ -3129,6 +3132,11 @@ impl<S> MidHandshakeSslStream<S> {
         self.stream.ssl()
     }
 
+    /// Returns a mutable reference to the `Ssl` of the stream.
+    pub fn ssl_mut(&mut self) -> &mut SslRef {
```
nox, line 3136: Exists to reset `TASK_CONTEXT_INDEX` on error.

boring/src/ssl/mod.rs
```diff
@@ -3383,6 +3391,11 @@ impl<S> SslStream<S> {
     pub fn ssl(&self) -> &SslRef {
         &self.ssl
     }
+
+    /// Returns a mutable reference to the `Ssl` object associated with this stream.
+    pub fn ssl_mut(&mut self) -> &mut SslRef {
```
nox, line 3396: Exists to reset `TASK_CONTEXT_INDEX` on error.

tokio-boring/src/async_callbacks.rs
```diff
@@ -0,0 +1,148 @@
...
+            + Send
+            + Sync
+            + 'static,
+    >,
+>;
+
+type BoxSelectCertFinish =
+    Box<dyn FnOnce(ClientHello<'_>) -> Result<(), AsyncSelectCertError> + 'static>;
+
+pub(crate) static TASK_CONTEXT_INDEX: Lazy<Index<Ssl, usize>> =
```
nox, line 20: I don't like how I use `usize` here but it's not that important, as other parts of boring already do to store a `SslContext` pointer in `Ssl`. We should probably use a pointer wrapper that implements `Send` and `Sync` for convenience at some point.

tokio-boring/src/async_callbacks.rs
```diff
@@ -0,0 +1,148 @@
...
+    Lazy::new(|| Ssl::new_ex_index().unwrap());
+pub(crate) static SELECT_CERT_FUTURE_INDEX: Lazy<Index<Ssl, BoxSelectCertFuture>> =
+    Lazy::new(|| Ssl::new_ex_index().unwrap());
+
+/// Extensions to [`SslContextBuilder`].
+///
+/// This trait provides additional methods to use async callbacks with boring.
+pub trait SslContextBuilderExt: private::Sealed {
```
nox, line 28: At first I thought I would just expose `async_select_certificate_callback` and `polling_select_certificate_callback` but using an extension trait makes usage more convenient.

tokio-boring/src/lib.rs
```diff
@@ -367,13 +385,18 @@ where
             stream: inner.stream,
             context: ctx as *mut _ as usize,
         };
+
         match (inner.f)(stream) {
             Ok(mut s) => {
                 s.get_mut().context = 0;
+                s.ssl_mut().set_ex_data(*TASK_CONTEXT_INDEX, 0);
```
nox, line 392: I do not like this. If there is a panic, neither of those will be reset to 0, which I think can lead to soundness issues afterwards, if you catch the unwind and try to use the boring API with the `SslContext` that uses async callbacks.

tokio-boring/src/lib.rs
```diff
@@ -367,13 +385,18 @@ where
...
+
                 Poll::Ready(Ok(StartedHandshake::Done(SslStream(s))))
             }
             Err(ssl::HandshakeError::WouldBlock(mut s)) => {
                 s.get_mut().context = 0;
+                s.ssl_mut().set_ex_data(*TASK_CONTEXT_INDEX, 0);
```
nox, line 398: Ditto soundness issue.

tokio-boring/src/lib.rs
```diff
@@ -396,13 +419,20 @@ where
         let mut s = self.0.take().expect("future polled after completion");
 
         s.get_mut().context = ctx as *mut _ as usize;
+        s.ssl_mut()
+            .set_ex_data(*TASK_CONTEXT_INDEX, ctx as *mut _ as usize);
+
         match s.handshake() {
             Ok(mut s) => {
                 s.get_mut().context = 0;
+                s.ssl_mut().set_ex_data(*TASK_CONTEXT_INDEX, 0);
```
nox, line 428: Ditto soundness issue.

tokio-boring/src/lib.rs
```diff
@@ -396,13 +419,20 @@ where
...
+
                 Poll::Ready(Ok(SslStream(s)))
             }
             Err(ssl::HandshakeError::WouldBlock(mut s)) => {
                 s.get_mut().context = 0;
+                s.ssl_mut().set_ex_data(*TASK_CONTEXT_INDEX, 0);
```
nox, line 434: Ditto soundness issue.

nox:
Oh oops I completely forgot I didn't yet implement setting `TASK_CONTEXT_INDEX` to anything useful.

boring/src/ssl/connector.rs
inikulin: What are your concerns?

nox: I managed to avoid needing this method.

tokio-boring/src/lib.rs
nox: Panics are segfaults in boring currently, added a note comment.

nox:
I removed the biggest breaking change (the removal of `accept` and `connect`).

nox:
The one missing thing from the PR is tests for `PrivateKeyMethod::decrypt` but I don't understand the circumstances that make boringssl invoke it.

boring/src/ssl/mod.rs
```diff
@@ -1958,15 +1987,19 @@ impl ClientHello {
...
+                ffi::SSL_early_callback_ctx_extension_get(self.0, ext_type.0, &mut ptr, &mut len);
             if result == 0 {
                 return None;
             }
             Some(slice::from_raw_parts(ptr, len))
         }
     }
 
-    fn ssl(&self) -> &SslRef {
+    pub fn ssl_mut(&mut self) -> &mut SslRef {
```
Noah-Kennedy, line 1998: Dangling pointer?

boring/src/ssl/mod.rs
```diff
@@ -1958,15 +1987,19 @@ impl ClientHello {
...
+        unsafe { SslRef::from_ptr_mut(self.0.ssl) }
+    }
+
+    pub fn ssl(&self) -> &SslRef {
```
Noah-Kennedy, line 2002: Dangling pointer?

tokio-boring/src/bridge.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// # Panics
+    ///
+    /// Panics if the bridge has no waker.
+    pub(crate) fn with_context<F, R>(&mut self, f: F) -> R
+    where
+        S: Unpin,
+        F: FnOnce(&mut Context<'_>, Pin<&mut S>) -> R,
+    {
+        let mut ctx =
+            Context::from_waker(self.waker.as_ref().expect("missing task context pointer"));
```
Noah-Kennedy, line 38: I would be very confused if I saw my program panic with this error, and would immediately go on completely the wrong path to debugging the issue.

Noah-Kennedy (commented):
I'm a bit nervous about a few things I see, and I'd like to go over this more thoroughly tomorrow or Monday.

boring/src/ssl/callbacks.rs
```diff
@@ -373,3 +367,93 @@ where
 
     callback(ssl, line);
 }
+
+pub(super) unsafe extern "C" fn raw_sign<M>(
```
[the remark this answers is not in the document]
nox: Yeah as a separate PR, none of the callback glue is documented.

boring/src/ssl/mod.rs
nox: Not sure what you mean with this question.

tokio-boring/src/bridge.rs
nox: Can you suggest an alternative e panic message?

boring/src/ssl/mod.rs
nox: Not sure what you mean with this question.

boring/src/ssl/mod.rs
nox: Not sure what you mean with this question.

boring/src/ssl/mod.rs
Noah-Kennedy: Nevermind, I misunderstood what this was doing internally. I didn't realize that the `from_ptr_mut` function returned `&'a mut Self`.

I assumed that this was a newtype over a smart pointer, and so I thought that a dangling ref to a smart pointer was being returned.

I see what you're doing here now, I agree that this is fine.

boring/src/ssl/mod.rs
Noah-Kennedy: Nevermind, I misunderstood what this was doing internally. I didn't realize that the `from_ptr_mut` function returned `&'a mut Self`.

I assumed that this was a newtype over a smart pointer, and so I thought that a dangling ref to a smart pointer was being returned.

I see what you're doing here now, I agree that this is fine.

tokio-boring/src/bridge.rs
Noah-Kennedy: "Internal error: missing waker in bridge"?

Noah-Kennedy: I suppose this is a bit of a nit.

Noah-Kennedy (approved):
@nox this looks good to go as far as I can tell.

Noah-Kennedy:
My remaining feedback is of the nit variety.

tokio-boring/src/async_callbacks.rs
```diff
@@ -0,0 +1,262 @@
...
+    /// itself a closure that will be passed [`ClientHello`] to configure
+    /// the connection based on the computations done in the future.
+    ///
+    /// See [`SslContextBuilder::set_select_certificate_callback`] for the sync
+    /// setter of this callback.
+    fn set_async_select_certificate_callback<Init, Fut, Finish>(&mut self, callback: Init)
+    where
+        Init: Fn(&mut ClientHello<'_>) -> Result<Fut, AsyncSelectCertError> + Send + Sync + 'static,
+        Fut: Future<Output = Result<Finish, AsyncSelectCertError>> + Send + Sync + 'static,
+        Finish: FnOnce(ClientHello<'_>) -> Result<(), AsyncSelectCertError> + 'static;
```
ghedo, line 48: Maybe I'm missing something, but I think the plan was to pass `&mut SslRef` instead of `ClientHello` to the closure? So we can actually change the various options on `SslRef`.

Also, not sure I understand why `Init` takes a `&mut ClientHello` rather than just `&ClientHello`? That doesn't need to be mutable.

nox: `ClientHello<'_>` and not `&mut SslRef` because there are methods that are only available on `ClientHello<'_>` (see the existing sync of the callback).

`&mut ClientHello<'_>` and not `&ClientHello<'_>` because getting a `&mut SslRef` from a `&ClientHello<'_>` is very much not good.

nox: Just to be extra clear: `ClientHello<'_>` is equipped with `fn ssl_mut(&mut self) -> &mut SslRef` in this PR.

ghedo: Oh, indeed I missed the `ssl_mut()` change, thanks. I understand now.

tokio-boring/src/bridge.rs
nox: I changed the message, using prefix "BUG:" like for all the other similar panic messages that exist in those crates.

Noah-Kennedy (approved):

boring/src/ssl/mod.rs
```diff
@@ -3526,60 +3603,68 @@ where
...
+    /// Initiates a server-side TLS handshake, returning a [`MidHandshakeSslStream`].
+    ///
+    /// This method calls [`Self::set_accept_state`] and returns without actually
+    /// initiating the handshake. The caller is then free to call
+    /// [`MidHandshakeSslStream`] and loop on [`HandshakeError::WouldBlock`].
+    pub fn setup_accept(mut self) -> MidHandshakeSslStream<S> {
+        self.set_accept_state();
 
         #[cfg(feature = "kx-safe-default")]
         stream.ssl.server_set_default_curves_list()?;
```
cjpatton, line 3646: CI is failing with this feature set: https://github.com/cloudflare/boring/actions/runs/6187838018/job/16798542713?pr=134

This ought to fix the issue, but you'll need setup_accept()` to return a `Result<_, Error>`.
```suggestion
        self.inner.ssl.server_set_default_curves_list()?;
```

Similarly for `setup_connect()`.

cjpatton (changes requested):

boring/src/ssl/mod.rs
nox: Ugh the whole point of adding those features was to make these infaillible 😭 I'll think about what to do.

cjpatton: :(

I think an argument can be made for `server_set_default_curves_list()` to panic instead of err if it fails. The value for the curvds list it's picking is hard-coded, and if we've built boringSSL properly it should work. In fact we have some unit tests that would excercise this panic.

nox: Good idea, I made `server_set_default_curves_list` and `client_set_default_curves_list` panic.

boring/src/ssl/test/mod.rs
```diff
@@ -1122,8 +1120,7 @@ fn client_set_default_curves_list() {
     let ssl_ctx = SslContextBuilder::new(SslMethod::tls()).unwrap().build();
     let mut ssl = Ssl::new(&ssl_ctx).unwrap();
 
-    ssl.client_set_default_curves_list()
-        .expect("Failed to set curves list. Is Kyber768 missing in boringSSL?")
```
cjpatton, line 1126: nit: Perhaps move "Is Kyber768 missing in boringSSL?" to a comment. Here and below.

cjpatton (approved):

boring/src/ssl/test/mod.rs
nox: Made it clear that it would panic if Kyber768 was missing, and I also put these changes in a proper commit prior to the introduction of `setup_accept` and `setup_connect`.

boring/src/ssl/mod.rs
```diff
@@ -3258,6 +3258,19 @@ impl SslRef {
     pub fn set_mtu(&mut self, mtu: u32) -> Result<(), ErrorStack> {
         unsafe { cvt(ffi::SSL_set_mtu(self.as_ptr(), mtu as c_uint) as c_int).map(|_| ()) }
     }
+
+    /// Sets the certificate.
+    ///
+    /// This corresponds to [`SSL_use_certificate`].
+    ///
+    /// [`SSL_use_certificate`]: https://www.openssl.org/docs/manmaster/man3/SSL_do_handshake.html
```
ghedo, line 3269: Wrong link,

nox: Good catch, I fixed it.

ghedo (approved):

PR #135, opened 2023-08-01
Some improvements and cleanups

nox:
Some stuff I wrote when working on #134, may as well make a separate PR.


boring/src/ssl/callbacks.rs
```diff
@@ -19,31 +21,35 @@ use crate::ssl::{
...
+    // SAFETY: boring provides valid inputs.
+    let ctx = unsafe { X509StoreContextRef::from_ptr_mut(x509_ctx) };
+
+    let ssl_idx = X509StoreContext::ssl_idx().expect("BUG: store context ssl index missing");
+    let verify_idx = SslContext::cached_ex_index::<F>();
+
+    let verify = ctx
+        .ex_data(ssl_idx)
+        .expect("BUG: store context missing ssl")
+        .ssl_context()
```
inikulin, line 40: Should we have `to_owned()` call here?

boring/src/ssl/callbacks.rs
```diff
@@ -95,73 +103,90 @@ where
...
-            Ok(psk_len) => psk_len as u32,
-            Err(e) => {
-                e.put();
-                0
-            }
+    // SAFETY: boring provides valid inputs.
+
+    let ssl = unsafe { SslRef::from_ptr_mut(ssl) };
+
+    let identity = if !identity.is_null() {
```
inikulin, line 110: nit: it's better avoid reverse conditions as it's easy to misread them or miss the negation during refactoring. So, let's return to the previous variant

boring/src/ssl/callbacks.rs
```diff
@@ -95,73 +103,90 @@ where
...
+        Some(unsafe { CStr::from_ptr(identity) }.to_bytes())
+    } else {
+        None
+    };
+
+    // Give the callback mutable slices into which it can write the psk.
+    let psk_sl = unsafe { slice::from_raw_parts_mut(psk as *mut u8, max_psk_len as usize) };
+
+    let ssl_context = ssl.ssl_context().to_owned();
```
inikulin, line 119: I think it worth adding a code comment describing why we call `to_owned` here and in other places. I can easily imagine it being removed during refactoring somewhere in the future.

I think we should expose internal `stored_callback` method on `ssl` that will under the hood obtain strong ref for ssl context and return it wrapped in `#[must_use]` struct along with fn pointer.

This way we somehow will have guarantees that callbacks are obtained in a correct way

boring/src/ssl/callbacks.rs
```diff
@@ -95,73 +103,90 @@ where
...
-    }
+    // SAFETY: boring provides valid inputs.
+    let ctx = unsafe { X509StoreContextRef::from_ptr_mut(x509_ctx) };
+
+    let ssl_idx = X509StoreContext::ssl_idx().expect("BUG: store context ssl index missing");
+
+    // NOTE(nox): I'm pretty sure this Arc<F> is unnecessary here as there is
+    // no way to get a `&mut SslRef` from a `&mut X509StoreContextRef`, and I
+    // don't understand how this callback is different from `raw_verify` above.
+    let callback = ctx
```
inikulin, line 148: shouldn't we obtain strong ref here?

inikulin (changes requested):

boring/src/ssl/callbacks.rs
nox: If we remove the `to_owned` calls, the code just doesn't type check anymore.

boring/src/ssl/callbacks.rs
nox: No, because the callback is owned as it is an `Arc<F>`. There is no unsafe code here to get `callback` so it wouldn't type check if it was incorrect anyway.

boring/src/ssl/callbacks.rs
nox: I explained just below in a `SAFETY` note why there is no need to own the context.

boring/src/ssl/callbacks.rs
nox: Ah yeah I copy-pasted that from `raw_client_psk` and then renamed `hint` to `authority`, I'll return to previous way here and change `raw_client_psk`.

inikulin (approved):

PR #138, opened 2023-08-03
Update RPK patch to latest patchset

rushilmehra:


tokio-boring/tests/rpk.rs
```diff
@@ -59,11 +57,12 @@ mod test_rpk {
 
         let client = async {
             let mut connector = SslConnector::rpk_builder().unwrap();
-            let cert = std::fs::read("tests/pubkey.der").unwrap();
 
-            connector.set_rpk_certificate(&cert).unwrap();
+            connector.set_raw_public_key_mode();
             let config = connector.build().configure().unwrap();
 
+            // allow client to configure custom verify callback w/ RPK
```
inikulin, line 64: need to implement that in test and provide a safe API if necessary

boring/src/ssl/mod.rs
```diff
@@ -712,25 +712,21 @@ impl SslContextBuilder {
...
-                cert.len() as u32,
-            ))
-            .map(|_| ())
+            ffi::SSL_CTX_set_raw_public_key_mode(self.as_ptr());
         }
     }
 
     /// Sets RPK null chain private key.
-    pub fn set_null_chain_private_key<T>(&mut self, key: &PKeyRef<T>) -> Result<(), ErrorStack>
+    pub fn set_raw_public_key_and_key<T>(&mut self, key: &PKeyRef<T>) -> Result<(), ErrorStack>
```
inikulin, line 722: `set_raw_public_key_and_key` -  the name suggest that this now also need to set a public key cert along with private one, no?

boring/src/ssl/mod.rs
```diff
@@ -3469,6 +3450,7 @@ where
         if ret > 0 {
             Ok(stream)
         } else {
+            println!("deez nuts error");
```
inikulin, line 3453: I think we don't want that in production 😄

boring/src/ssl/mod.rs
rushilmehra: the public key is automatically generated from the private key and stored on SSL_CTX. https://boringssl-review.googlesource.com/c/boringssl/+/45444/18 see docs on the function header

boring/src/ssl/mod.rs
rushilmehra: lmao whoops

boring/src/ssl/callbacks.rs
```diff
@@ -48,6 +48,30 @@ where
...
 }
 
+pub(super) unsafe extern "C" fn raw_custom_verify<F>(
+    ssl: *mut ffi::SSL,
+    al: *mut u8,
+) -> ffi::ssl_verify_result_t
+where
+    F: Fn(&mut SslRef, &mut u8) -> bool + 'static + Sync + Send,
+{
+    unsafe {
```
inikulin, line 58: Let's limit unsafe block scope only to expressions that are actually unsafe

boring/src/ssl/callbacks.rs
```diff
@@ -48,6 +48,30 @@ where
     verify(preverify_ok != 0, ctx) as c_int
 }
 
+pub(super) unsafe extern "C" fn raw_custom_verify<F>(
+    ssl: *mut ffi::SSL,
+    al: *mut u8,
+) -> ffi::ssl_verify_result_t
+where
+    F: Fn(&mut SslRef, &mut u8) -> bool + 'static + Sync + Send,
```
inikulin, line 56: Worth making API safer here by making the closure accept a mut ref to https://docs.rs/boring/latest/boring/ssl/struct.SslAlert.html instead of raw u8

boring/src/ssl/mod.rs
```diff
@@ -703,31 +696,41 @@ impl SslContextBuilder {
...
 
-    /// Sets raw public key certificate in DER format.
-    pub fn set_rpk_certificate(&mut self, cert: &[u8]) -> Result<(), ErrorStack> {
+    /// SSL_CTX_set_raw_public_key_mode indicates that |ctx| wishes to receive a raw
+    /// public key from a server rather than an X.509 certificate chain. Clients
+    /// should install a custom verifier with |SSL_CTX_set_custom_verify| or
+    /// |SSL_set_custom_verify| to verify the server's public key, which can be
+    /// obtained via |SSL_get0_peer_pubkey| or |SSL_get0_unparsed_peer_pubkey|. This
+    /// is only supported for client connections currently and only with TLSv1.3.
+    pub fn set_raw_public_key_mode(&mut self) {
```
inikulin, line 713: We shouldn't expose as the only right way to construct RPK is via Builde::new_rpk - it also ensures that RPK index var is set that is used to assert that RPK-incompatible API is not used

boring/src/ssl/mod.rs
```diff
@@ -703,31 +696,41 @@ impl SslContextBuilder {
...
-    /// Sets RPK null chain private key.
-    pub fn set_null_chain_private_key<T>(&mut self, key: &PKeyRef<T>) -> Result<(), ErrorStack>
+    /// SSL_CTX_set_raw_public_key_and_key sets a raw public key (in
+    /// SubjectPublicKeyInfo format) and the corresponding private key for a TLS
+    /// server. References to the given |CRYPTO_BUFFER| and |EVP_PKEY| objects are
+    /// added as needed. Exactly one of |privkey| or |privkey_method| may be
+    /// non-NULL. |spki| must be non-NULL iff |privkey_method| is, otherwise the
+    /// public key will be generated from |privkey|. It returns one on success and
+    /// zero on error.
+    pub fn set_raw_public_key_and_key<T>(&mut self, key: &PKeyRef<T>) -> Result<(), ErrorStack>
```
inikulin, line 726: We need to mention here that this need to be used along with `set_custom_verify`. In addition we should introduce a convenience method that automatically creates a pub key from the provided private key and sets up a callback like you do in tests.

inikulin (changes requested):

boring/src/ssl/mod.rs
```diff
@@ -703,37 +696,49 @@ impl SslContextBuilder {
...
+    /// SubjectPublicKeyInfo format) and the corresponding private key for a TLS
+    /// server. References to the given `CRYPTO_BUFFER` and `EVP_PKEY` objects are
+    /// added as needed. Exactly one of `privkey` or `privkey_method` may be
+    /// non-NULL. `spki` must be non-NULL iff `privkey_method` is, otherwise the
+    /// public key will be generated from `privkey`. It returns one on success and
+    /// zero on error. Clients should install a custom verifier with
+    /// `SSL_CTX_set_custom_verify` or `SSL_set_custom_verify` to verify the server's
+    /// public key, which can be obtained via `SSL_get0_peer_pubkey` or
+    /// `SSL_get0_unparsed_peer_pubkey`. This is only supported for client connections
+    /// currently and only with TLSv1.3.
```
nox, line 715: Let's not just copy-paste the docs from boringssl, most of this stuff is only relevant in the context of C code.

boring/src/ssl/mod.rs
```diff
@@ -805,6 +810,25 @@ impl SslContextBuilder {
         }
     }
 
+    /// Configures the certificate verification method for new connections.
+    ///
+    /// This corresponds to [`SSL_CTX_set_cusom_verify`].
```
nox, line 815:
```suggestion
    /// This corresponds to [`SSL_CTX_set_custom_verify`].
```

boring/src/ssl/mod.rs
rushilmehra: fixed

boring/src/ssl/mod.rs
rushilmehra: fixed, thanks

boring/src/ssl/mod.rs
```diff
@@ -709,37 +702,41 @@ impl SslContextBuilder {
...
                 self.as_ptr(),
+                ptr::null_mut(),
                 key.as_ptr(),
                 ptr::null_mut(),
             ))
             .map(|_| ())
         }
     }
+
+    pub fn configure_default_rpk_custom_verify(&mut self, cert: &[u8]) -> Result<(), ErrorStack> {
```
nox, line 729: This is undocumented.

boring/src/ssl/mod.rs
```diff
@@ -811,6 +808,25 @@ impl SslContextBuilder {
...
     }
 
+    /// Configures the certificate verification method for new connections.
+    ///
+    /// This corresponds to [`SSL_CTX_set_custom_verify`].
+    ///
+    /// [`SSL_CTX_set_custom_verify`]: https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_CTX_set_custom_verify
+    pub fn set_custom_verify<F>(&mut self, mode: SslVerifyMode, verify: F)
+    where
+        F: Fn(&mut SslRef, &mut u8) -> bool + 'static + Sync + Send,
```
nox, line 818: `&mut SslAlert`

nox (changes requested):

PR #139, opened 2023-08-03
Fix feature gates for fips-link-precompiled. Cut a new release

inikulin:


Cargo.toml
```diff
@@ -8,7 +8,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
-version = "3.0.1"
+version = "3.0.2"
```
nox, line 11: Tbh I'm still worried about the workspace version number, it makes it hard to track what is supposed to be released etc. E.g. a breaking change (or any change actually) in tokio-boring shouldn't lead to a bump (breaking or not) of boring and boring-sys. Could we reconsider this decision please?

nox (approved):

Cargo.toml
inikulin: copying from internal chat for visibility:

for version numbers - all those crates are usually used in conjunction with each other. And considering the nature of the sys crate (linking with native lib doesn't allow proper semver) it's much easier to use with versions being pinned. 

What are potential problematic scenarios you see with this approach?

also, I have no idea how to track which crates have changed and which didn't between releases. Assessing that manually is extremely error-prone.

Also, note that we need to do major bump every time we do a bump of borinssl revision, having disjointed versions it will be extremely hard to figure out crates compatibility. E.g. hyper-boring 5.0.0 would require boring 3.0.0

PR #145, opened 2023-08-09
Use features to set key exchange preferences

cjpatton:
Overwrite boringSSL's default key exchange preferences with safe
defaults using feature flags:

* "kx-client-pq-supported" enables support for PQ key exchange algorithms.
  Classical key exchange is still preferred, but will be upgraded to PQ
  if requested.

* "kx-client-pq-preferred" enables preference for PQ key exchange,
  with fallback to classical key exchange if requested.

* "kx-client-nist-required" disables non-FIPS-compliant algorithms on the client side.

If any of these "kx-*" features are enabled, then don't compile bindings
for `SSL_CTX_set1_curves()`. This is to prevent the feature flags from
silently overriding curve preferences chosen by the user.

Ideally we'd allow both: that is, use "kx-\*" to set defaults, but still
allow the user to manually override them. However, this doesn't work
because by the time the `SSL_CTX` is constructed, we don't yet know
whether we're the client or server. (The "kx-\*" features set different
preferences for each.)

### DIFF
--- a/boring/Cargo.toml
+++ b/boring/Cargo.toml
@@ -16,6 +16,8 @@ features = ["rpk", "pq-experimental"]
 rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
+# Controlling the build
+
 # Use a FIPS-validated version of boringssl.
 fips = ["boring-sys/fips"]
 
@@ -25,16 +27,42 @@ fips-link-precompiled = ["boring-sys/fips-link-precompiled"]
 # Enables Raw public key API (https://datatracker.ietf.org/doc/html/rfc7250)
 rpk = ["boring-sys/rpk"]
 
-# Enables experimental post-quantum crypto (https://blog.cloudflare.com/post-quantum-for-all/)
+# Applies a patch to the boringSSL source code that enables support for PQ key
+# exchange. This feature is necessary in order to compile the bindings for the
+# default branch of boringSSL. Alternatively, a version of boringSSL that
+# implements the same feature set can be provided by setting
+# `BORING_BSSL_SOURCE_PATH`.
 pq-experimental = ["boring-sys/pq-experimental"]
 
 # Disables git patching of the BoringSSL sources for features like `rpk` and `pq-experimental`, but
-# keeps the related Rust API. 
-# 
-# Supposed to be used with either pre-compiled BoringSSL (via `BORING_BSSL_PATH` env variable) or 
-# with custom BoringSSL sources (via `BORING_BSSL_SOURCE_PATH` env variable) already containing 
+# keeps the related Rust API.
+#
+# Supposed to be used with either pre-compiled BoringSSL (via `BORING_BSSL_PATH` env variable) or
+# with custom BoringSSL sources (via `BORING_BSSL_SOURCE_PATH` env variable) already containing
 # required patches.
 no-patches = ["boring-sys/no-patches"]
+
+# Controlling key exchange preferences at compile time
+
+# Choose key exchange preferences at compile time. This prevents the user from
+# choosing their own preferences.
+kx-safe-default = []
+
+# Support PQ key exchange. The client will prefer classical key exchange, but
+# will upgrade to PQ key exchange if requested by the server. This is the
+# safest option if you don't know if the peer supports PQ key exchange. This
+# feature implies "kx-safe-default".
+kx-client-pq-supported = ["kx-safe-default"]
+
+# Prefer PQ key exchange. The client will prefer PQ exchange, but fallback to
+# classical key exchange if requested by the server. This is the best option if
+# you know the peer supports PQ key exchange. This feature implies
+# "kx-safe-default" and "kx-client-pq-supported".
+kx-client-pq-preferred = ["kx-safe-default", "kx-client-pq-supported"]
+
+# Disable key exchange involving non-NIST key exchange on the client side.
+# Implies "kx-safe-default".
+kx-client-nist-required = ["kx-safe-default"]
 
 [dependencies]
 bitflags = { workspace = true }
--- a/boring/src/ssl/mod.rs
+++ b/boring/src/ssl/mod.rs
@@ -619,10 +619,12 @@ impl SslSignatureAlgorithm {
 }
 
 /// A TLS Curve.
+#[cfg(not(feature = "kx-safe-default"))]
 #[repr(transparent)]
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub struct SslCurve(c_int);
 
+#[cfg(not(feature = "kx-safe-default"))]
 impl SslCurve {
     pub const SECP224R1: SslCurve = SslCurve(ffi::NID_secp224r1);
 
@@ -1703,6 +1705,11 @@ impl SslContextBuilder {
     /// This corresponds to [`SSL_CTX_set1_curves`]
     ///
     /// [`SSL_CTX_set1_curves`]: https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_CTX_set1_curves
+    //
+    // If the "kx-*" flags are used to set key exchange preference, then don't allow the user to
+    // set them here. This ensures we don't override the user's preference without telling them:
+    // when the flags are used, the preferences are set just before connecting or accepting.
+    #[cfg(not(feature = "kx-safe-default"))]
     pub fn set_curves(&mut self, curves: &[SslCurve]) -> Result<(), ErrorStack> {
         unsafe {
             cvt_0i(ffi::SSL_CTX_set1_curves(
@@ -2398,6 +2405,48 @@ impl SslRef {
 
     fn get_error(&self, ret: c_int) -> ErrorCode {
         unsafe { ErrorCode::from_raw(ffi::SSL_get_error(self.as_ptr(), ret)) }
+    }
+
+    #[cfg(feature = "kx-safe-default")]
+    fn set_curves_list(&mut self, curves: &str) -> Result<(), ErrorStack> {
+        let curves = CString::new(curves).unwrap();
+        unsafe {
+            cvt(ffi::SSL_set1_curves_list(
+                self.as_ptr(),
+                curves.as_ptr() as *const _,
+            ))
+            .map(|_| ())
+        }
+    }
+
+    #[cfg(feature = "kx-safe-default")]
+    fn client_set_default_curves_list(&mut self) -> Result<(), ErrorStack> {
+        let curves = if cfg!(feature = "kx-client-pq-preferred") {
+            if cfg!(feature = "kx-client-nist-required") {
+                "P256Kyber768Draft00:P-256:P-384"
+            } else {
+                "X25519Kyber768Draft00:X25519:P256Kyber768Draft00:P-256:P-384"
+            }
+        } else if cfg!(feature = "kx-client-pq-supported") {
+            if cfg!(feature = "kx-client-nist-required") {
+                "P-256:P-384:P256Kyber768Draft00"
+            } else {
+                "X25519:P-256:P-384:X25519Kyber768Draft00:P256Kyber768Draft00"
+            }
+        } else {
+            if cfg!(feature = "kx-client-nist-required") {
+                "P-256:P-384"
+            } else {
+                "X25519:P-256:P-384"
+            }
+        };
+
+        self.set_curves_list(curves)
+    }
+
+    #[cfg(feature = "kx-safe-default")]
+    fn server_set_default_curves_list(&mut self) -> Result<(), ErrorStack> {
+        self.set_curves_list("X25519Kyber768Draft00:P256Kyber768Draft00:X25519:P-256:P-384")
     }
 
     /// Like [`SslContextBuilder::set_verify`].
@@ -3480,6 +3529,10 @@ where
     /// See `Ssl::connect`
     pub fn connect(self) -> Result<SslStream<S>, HandshakeError<S>> {
         let mut stream = self.inner;
+
+        #[cfg(feature = "kx-safe-default")]
+        stream.ssl.client_set_default_curves_list()?;
+
         let ret = unsafe { ffi::SSL_connect(stream.ssl.as_ptr()) };
         if ret > 0 {
             Ok(stream)
@@ -3503,6 +3556,10 @@ where
     /// See `Ssl::accept`
     pub fn accept(self) -> Result<SslStream<S>, HandshakeError<S>> {
         let mut stream = self.inner;
+
+        #[cfg(feature = "kx-safe-default")]
+        stream.ssl.server_set_default_curves_list()?;
+
         let ret = unsafe { ffi::SSL_accept(stream.ssl.as_ptr()) };
         if ret > 0 {
             Ok(stream)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -238,3 +238,5 @@ jobs:
       name: Run `pq-experimental` tests
     - run: cargo test --features pq-experimental,rpk
       name: Run `pq-experimental,rpk` tests
+    - run: cargo test --features kx-safe-default,pq-experimental
+      name: Run `kx-safe-default` tests
--- a/boring-sys/Cargo.toml
+++ b/boring-sys/Cargo.toml
@@ -63,7 +63,11 @@ fips-link-precompiled = []
 # Enables Raw public key API (https://datatracker.ietf.org/doc/html/rfc7250)
 rpk = []
 
-# Enables experimental post-quantum crypto (https://blog.cloudflare.com/post-quantum-for-all/)
+# Applies a patch (`patches/boring-pq.patch`) to the boringSSL source code that
+# enables support for PQ key exchange. This feature is necessary in order to
+# compile the bindings for the default branch of boringSSL (`deps/boringssl`).
+# Alternatively, a version of boringSSL that implements the same feature set
+# can be provided by setting `BORING_BSSL_SOURCE_PATH`.
 pq-experimental = []
 
 # Disables git patching of the BoringSSL sources for features like `rpk` and `pq-experimental`, but
--- a/boring/src/ssl/test/mod.rs
+++ b/boring/src/ssl/test/mod.rs
@@ -1115,3 +1115,23 @@ fn session_cache_size() {
     let ctx = ctx.build();
     assert_eq!(ctx.session_cache_size(), 1234);
 }
+
+#[cfg(feature = "kx-safe-default")]
+#[test]
+fn client_set_default_curves_list() {
+    let ssl_ctx = SslContextBuilder::new(SslMethod::tls()).unwrap().build();
+    let mut ssl = Ssl::new(&ssl_ctx).unwrap();
+
+    ssl.client_set_default_curves_list()
+        .expect("Failed to set curves list. Is Kyber768 missing in boringSSL?")
+}
+
+#[cfg(feature = "kx-safe-default")]
+#[test]
+fn server_set_default_curves_list() {
+    let ssl_ctx = SslContextBuilder::new(SslMethod::tls()).unwrap().build();
+    let mut ssl = Ssl::new(&ssl_ctx).unwrap();
+
+    ssl.server_set_default_curves_list()
+        .expect("Failed to set curves list. Is Kyber768 missing in boringSSL?")
+}


boring/Cargo.toml
```diff
@@ -16,10 +16,12 @@ features = ["rpk", "pq-experimental"]
 rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
-# Use a FIPS-validated version of boringssl.
+# Use a FIPS-validated version of boringssl. Moreover, only support the NIST
+# curves for the key exchange by default.
 fips = ["boring-sys/fips"]
```
cjpatton, line 21: Question for reviewer: "fips" and "rpk" are meant to be mutually exclusive. What about "fips" and "pq-experimental"?

boring/src/ssl/mod.rs
```diff
@@ -692,17 +728,38 @@ pub struct SslContextBuilder {
     is_rpk: bool,
 }
 
+unsafe fn new_ssl_ctx_with_default_curves(
+    method: SslMethod,
+) -> Result<*mut ffi::SSL_CTX, ErrorStack> {
+    init();
+    let ctx = cvt_p(ffi::SSL_CTX_new(method.as_ptr()))?;
+    ssl_ctx_set_curves(ctx, SslCurve::DEFAULT_CURVES)?;
```
inikulin, line 736: instead of using ffi, we can use safe API on builder: https://docs.rs/boring/latest/boring/ssl/struct.SslContextBuilder.html#method.set_curves

boring/src/ssl/mod.rs
```diff
@@ -641,6 +641,42 @@ impl SslCurve {
...
+                SslCurve::P256_KYBER768_DRAFT00,
+            ];
+        } else {
+            const DEFAULT_CURVES: &[SslCurve] = &[
+                #[cfg(not(feature = "fips"))]
+                #[cfg(not(feature = "fips-link-precompiled"))]
+                SslCurve::X25519,
+                SslCurve::SECP256R1,
+                SslCurve::SECP384R1,
+            ];
```
inikulin, line 677: To not introduce breaking changes, let's set the default curves only if any of the features is enabled

inikulin (changes requested):

boring/src/ssl/mod.rs
cjpatton: That makes sense, but one of the goals of this PR is to disable X25519 if the `fips` flag is set. Do you agree with this goal?

Another way to achieve the same effect would be to call `SSL_CTX_set_compliance_policy()`, which configures P256 and P384 only: https://boringssl.googlesource.com/boringssl/+/refs/heads/master/include/openssl/ssl.h#5326. If you agree with this goal, then let's do this instead?

boring/src/ssl/mod.rs
cjpatton: So the reason I'm using the unsafe API here is that I want to call `SSL_CTX_set1_curves()` wherever we call `SSL_CTX_new()`. If you look at the constructors for `SslContextBuilder`, they all use the unsafe API. I coiuld make `new_ssl_ctx_with_default_curves()` safe, but that would require more wrangling because `SslContextBuilder` also has the `is_rpk` flag.

Also, note that `ssl_ctx_set_curves()` is now also used by `SslContextBuilder::set_curves()`.

boring/Cargo.toml
inikulin: iirc according to @bwesterb it should be compatible

boring/src/ssl/mod.rs
inikulin: E.g. instead of:
```rust 
    pub fn new_rpk() -> Result<SslContextBuilder, ErrorStack> {
        unsafe {
            init();
            let ctx = cvt_p(ffi::SSL_CTX_new(SslMethod::tls_with_buffer().as_ptr()))?;

            Ok(SslContextBuilder::from_ptr(ctx, true))
        }
    }
```
    have
```rust
    pub fn new_rpk() -> Result<SslContextBuilder, ErrorStack> {
        let mut ctx = unsafe {
            init();
            let ctx = cvt_p(ffi::SSL_CTX_new(SslMethod::tls_with_buffer().as_ptr()))?;

            SslContextBuilder::from_ptr(ctx, true)
        };

       ctx.set_curves(SslCurve::DEFAULT_CURVES)?;

        Ok(ctx)
    }
```

boring/Cargo.toml
cjpatton: Hmm, I ask because I wonder if there is a technical reason for the code below that would also apply to the PQ patch (from boring-sys/build.rs):
```rust
    #[cfg(all(feature = "fips", feature = "rpk"))]
    compile_error!("`fips` and `rpk` features are mutually exclusive");
```

cjpatton: After discussing with Rushil offline, it turns out that "fips" is mutually exclusive with any feature that depends on a newer version of boringSSL or the patches.

rushilmehra: @bwesterb My understanding was that  the "fips" feature implies we just use the latest FIPS validated commit from the GH mirror and don't mess with it. Do we apply the PQ patch on top of that?

boring/src/ssl/mod.rs
cjpatton: It turns out that some of the tests related to SRTP fail if this compliance policy is set. I've fallen back to just explicitly enabling FIPS only key exchanges.

boring/Cargo.toml
cjpatton: Based on https://github.com/cloudflare/boring/blob/master/boring-sys/build.rs#L120-L123, it looks to me like `boringssl-fips` is used if "fips" is set and "fips-link-precompiled" is not set. Otherwise, `boringssl` is used.

The PQ patch applies to `boringssl`, but likely not `boringssl-fips` since this version of boringSSL seems to predate Kyber.

So "pq-experimental,fips-link-precompiled" should build, but "pq-experimental,fips" probably doesn't.

cjpatton:
I've updated the commit based on feedback on this PR and clarification of the semantics of "fips" and "fips-link-compiled". Please take another look!

boring/Cargo.toml
```diff
@@ -16,20 +16,38 @@ features = ["rpk", "pq-experimental"]
...
-# Use a FIPS-validated version of boringssl.
+# Use the FIPS-validated version of boringSSL. This feature is incompatible
+# with any feature that requires (patches to) the latest version of boringSSL.
 fips = ["boring-sys/fips"]
 
-# Link with precompiled FIPS-validated `bcm.o` module.
-fips-link-precompiled = ["fips", "boring-sys/fips-link-precompiled"]
+# Link with a precompiled FIPS-validated `bcm.o` module in order to use the
+# FIPS-validated implementations of key exchange and encryption algorithms.
+fips-link-precompiled = ["boring-sys/fips-link-precompiled"]
```
inikulin, line 25: This won't work: we enabled `fips` here to enable safe API that's under `fips` flag and becomes available when we link precompiled. 

If we want to separate the two we would need to go over the safe code and replace `cfg(feature = "fips")` with `cfg(any(feature = "fips", feature = "fips-link-precompiled"))`. 

Same applies to the build script: places where we had `cfg(all(not(feature = "fips"), feature = "fips-link-precompiled")` need to be replaced with just `cfg(feature = "fips-link-precompiled")` and places where we have just `cfg(feature = "fips")` need to be changed to `cfg(any(feature = "fips", feature = "fips-link-precompiled"))`

boring-sys/build.rs
```diff
@@ -584,7 +584,19 @@ fn main() {
...
+    // exclusive. Furthermore, any feature that requires a newer version of boringSSL, or patching
+    // boringSSL, is incompatible with "fips".
+    #[cfg(all(
+        feature = "fips",
+        any(
+            feature = "fips-link-precompiled",
+            feature = "rpk",
+            feature = "pq-experimental"
+        )
+    ))]
```
inikulin, line 599: Compile error message needs an update

boring/src/ssl/mod.rs
```diff
@@ -1707,6 +1715,39 @@ impl SslContextBuilder {
...
+                self.set_curves(&[
+                    #[cfg(not(feature = "fips-required"))]
+                    SslCurve::X25519_KYBER512_DRAFT00,
+                    SslCurve::P256_KYBER768_DRAFT00,
+                    #[cfg(not(feature = "fips-required"))]
+                    SslCurve::X25519,
+                    SslCurve::SECP256R1,
+                    SslCurve::SECP384R1,
+                ])?;
+            } else if #[cfg(feature = "pq-supported")] {
```
inikulin, line 1730: Does that mean `pq-preferred` and `pq-supported` are mutually exclusive?

boring/src/ssl/mod.rs
```diff
@@ -696,16 +696,20 @@ pub struct SslContextBuilder {
 impl SslContextBuilder {
     /// Creates a new `SslContextBuilder` to be used with Raw Public Key.
     ///
-    /// This corresponds to [`SSL_CTX_new`].
+    /// This corresponds to calling [`SSL_CTX_new`] then setting the default key exchange
+    /// preferences.
```
inikulin, line 700: Worth updating code comments with information that default curves will be changed based on the featureset

boring/Cargo.toml
```diff
@@ -16,20 +16,38 @@ features = ["rpk", "pq-experimental"]
...
+# Support PQ key exchange by default. The client will prefer classical key
+# exchange, but will upgrade to PQ key exchange if requested by the server.
+# This is the safest option if you don't know if the peer supports PQ key
+# exchange.
+pq-supported = ["pq-experimental"]
+
+# Prefer PQ key exchange by default. The client will prefer PQ exchange, but
+# fallback to classical key exchange if requested by the server. This is the
+# best option if you know the peer supports PQ key exchange.
+pq-preferred = ["pq-experimental", "pq-supported"]
```
inikulin, line 42: `pq-preferred` is confusing as one might think it's a preferred configuration, rather than PQ exchange being preferred, let's try to reflect the meaning in feature names

inikulin (changes requested):

boring/Cargo.toml
cjpatton: Hmm, so the reason I made this change is that I noted the following when I try to build with "fips-link-precompiled" (this is reproducible on `master`):


```
chris@ubutron:~/github.com/cjpatton/boring$ export RUSTFLAGS="-Clink-args=-Wl,-zmuldefs" && export BORING_SSL_PRECOMPILED_BCM_O=/opt/boringssl-fips/lib/bcm.o && cargo clean && cargo test --features fips-link-precompiled
                ... snipped ...
   Compiling tokio v1.30.0
   Compiling boring-sys v3.0.4 (/home/chris/github.com/cjpatton/boring/boring-sys)
   Compiling rusty-hook v0.11.2
   Compiling futures-executor v0.3.28
   Compiling futures v0.3.28
   Compiling tokio-util v0.7.8
   Compiling h2 v0.3.20
   Compiling hyper v0.14.27
warning: linking in precompiled bcm.c.o` module
   Compiling boring v3.0.4 (/home/chris/github.com/cjpatton/boring/boring)
error: could not find native static library `crypto`, perhaps an -L flag is missing?

The following warnings were emitted during compilation:

warning: linking in precompiled bcm.c.o` module

error: could not compile `boring-sys` (lib) due to previous error
warning: build failed, waiting for other jobs to finish...
error[E0308]: mismatched types
    --> boring/src/bio.rs:33:17
     |
31   |             cvt_p(BIO_new_mem_buf(
     |                   --------------- arguments to this function are incorrect
32   |                 buf.as_ptr() as *const _,
33   |                 buf.len() as BufLen,
     |                 ^^^^^^^^^^^^^^^^^^^ expected `isize`, found `i32`
     |
note: function defined here
    --> /home/chris/github.com/cjpatton/boring/target/debug/build/boring-sys-f0b0def01807e627/out/bindings.rs:9461:12
     |
9461 |     pub fn BIO_new_mem_buf(buf: *const ::std::os::raw::c_void, len: ossl_ssize_t) -> *mut BIO;
     |            ^^^^^^^^^^^^^^^
help: you can convert an `i32` to an `isize` and panic if the converted value doesn't fit
     |
33   |                 (buf.len() as BufLen).try_into().unwrap(),
     |                 +                   +++++++++++++++++++++

error[E0308]: mismatched types
     --> boring/src/ssl/mod.rs:1316:17
      |
1313  |             let r = ffi::SSL_CTX_set_alpn_protos(
      |                     ---------------------------- arguments to this function are incorrect
...
1316  |                 protocols.len() as ProtosLen,
      |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `usize`, found `u32`
      |
                       ... snipped ...
```

cjpatton: This happens because the rust bindings are no longer compatible with `ffi` with the version of boring in boring-sys/dep/boringssl-fips.

boring/src/ssl/mod.rs
cjpatton: `pq-preferred` implies `pq-supported`: https://github.com/cjpatton/boring/blob/pq-by-default/boring/Cargo.toml#L42

So if `pq-preferred` is set but `pq-supported` is, the `else if` will be taken here. Otherwise, if `pq-preferred` is set, then the `if` above will be taken. (This is what we want.)

boring/src/ssl/mod.rs
cjpatton: Yup! Done.

boring/Cargo.toml
cjpatton: What about "pq-key-exchange-preferred"?

boring-sys/build.rs
cjpatton: Good catch, updated. Though I may be misunderstnding how things work, I think I need to do more digging to see how the "fips-link-precompiled" and "fips" are supposed to interact. (I assumed they were mutually exclusive but perhaps not?)

boring/Cargo.toml
cjpatton: See #147.

boring/Cargo.toml
cjpatton: Changed to "pq-kx-preferred".

boring/src/ssl/mod.rs
```diff
@@ -1707,6 +1732,33 @@ impl SslContextBuilder {
         }
     }
 
+    fn set_safe_defaults(&mut self) -> Result<(), ErrorStack> {
+        // Choose a curves list suitable for the given feature set.
```
cjpatton, line 1736: @bwesterb can you provide some guidance here? What do you think would be make sensible defaults?

boring/src/ssl/mod.rs
```diff
@@ -1177,6 +1185,23 @@ impl SslContextBuilder {
         }
     }
 
+    /// Sets the list preferred curves for the context to be the colon-separated list |curves|.
+    /// Each element of `curves` should be a curve name (e.g. P-256, X25519, ...).
```
bwesterb, line 1189: Nit: I'd just give a concrete example; eg. `P-256:X25519`.

boring/src/ssl/mod.rs
```diff
@@ -1707,6 +1732,33 @@ impl SslContextBuilder {
         }
     }
 
+    fn set_safe_defaults(&mut self) -> Result<(), ErrorStack> {
+        // Choose a curves list suitable for the given feature set.
+        let curves = if cfg!(feature = "kx-pq-preferred") {
+            if cfg!(feature = "kx-fips-required") {
+                Some("P256Kyber768Draft00:P-256:P-384")
+            } else {
+                Some("X25519Kyber768Draft00:X25519:P256Kyber768Draft00:P-256:P-384")
```
bwesterb, line 1741: We need to prefer PQ over non-PQ, so:
```
X25519Kyber768Draft00:P256Kyber768Draft00:X25519:P-256:P-384
```

boring/src/ssl/mod.rs
```diff
@@ -1707,6 +1732,33 @@ impl SslContextBuilder {
...
+            }
+        } else if cfg!(feature = "kx-pq-supported") {
+            if cfg!(feature = "kx-fips-required") {
+                Some("P-256:P-384:P256Kyber768Draft00")
```
bwesterb, line 1745: This is fine if it's client-side, but not if it's server-side. For server-side we want to have PQ in the front.

I think this sets the defaults for both client and server side?

boring/src/ssl/mod.rs
```diff
@@ -1707,6 +1732,33 @@ impl SslContextBuilder {
...
+            } else {
+                Some("X25519:P-256:P-384:X25519Kyber768Draft00:P256Kyber768Draft00")
```
bwesterb, line 1747: Same applies here.

boring/src/ssl/mod.rs
cjpatton: Good idea, done. FWIW this paragraph was copied from boringSSL's API docs :D

boring/src/ssl/mod.rs
bwesterb: Many services act as both client and server, so the current approach will not work.

cjpatton: The problem is that the client and server need to have different default curve preferences. Options:
1. Remove `kx-fips-required` flag and force clients to call `SslContextBuilder::set_curves_list()` manually.
2. Wait to resolve default preferences until `Ssl::accept()` or `Ssl::connect()` is called. Background: The `SslContext` does not know whether it will be used for the client or server. This is partially determined by `SslMethod`, but there are methods that do not specify the role. What we could do instead is propagate a flag from `SslContextBuilder` to `SslContext` to `Ssl` that indicates whether the default curve preferences have been overwritten by the user, and if not, set the curves list just before `accept` or `connect`.

The second option is preferable because it requires no intervention on the part of the user.

cjpatton: Option 2 is pretty tricky as it turns out! I considered a few options for adding the flag:

2(a) Wrap `SslContext`, i.e., `pub struct WrappedSslContext{ inner: SslContext, curve_preference_set_by_user: bool }`. This is the safest option because we don't have to mess with the FFI bits. However, this requires a major refactor (most consumers of `SslContext` expect a thin wrapper around `SSL_CTX`).

2(b) Change the type wrapped by `SslContext` define a C struct that contains the `SSL_CTX` and the flag. This is not ideal because it requires doing surgery on `unsafe` code.

3(b) Create a macro similar to `generic_foreign_type_and_impl_send_sync!` that wraps a foreign type _and_ and a rust struct, perhaps one that is `Default` and `Copy` to keep things simple. The details are a bit tricky, but this seems promising and I'm still working on it.

A couple of other options I wanted to flag, which would be simpler to implement:

(4) Patch boringSSL with a method that checks if the user changed the default curve preferences. This is not ideal because the implementation would subject to change as boringSSL evolves.

(5) When the feature flags are set **disable** manually setting curve preferences, i.e., set_curve_list and set_curves would return an error if any of "kx-pq-preferred", "kx-fips-required", etc. are used. This is easy to implement but is more restrictive than we'd like.

cjpatton: I opted for (5) due to its simplicity and fail safety: https://github.com/cloudflare/boring/pull/145/commits/9eedf6178afda50703f4e1e1b07a15b76586a68e

bwesterb: I think (5) is elegantly pragmatic. If we have a user that needs to set curve preferences, we can change to something like (4): IIRC BoringSSL uses an empty list if curve preferences aren't set.

bwesterb (approved):
This is alright if we have a BoringSSL that disables "second keyshare" by default.

boring/Cargo.toml
```diff
@@ -25,9 +25,26 @@ fips-link-precompiled = ["fips", "boring-sys/fips-link-precompiled"]
...
+# safest option if you don't know if the peer supports PQ key exchange.
+kx-pq-supported = []
+
+# Prefer PQ key exchange. The client will prefer PQ exchange, but fallback to
+# classical key exchange if requested by the server. This is the best option if
+# you know the peer supports PQ key exchange.
+kx-pq-preferred = ["kx-pq-supported"]
+
+# Disable non-FIPS key exchange algorithms on the client side.
+kx-fips-required = []
```
cjpatton, line 46: By "FIPS" here I mean the algorithms exported by the FIPS bcm.o. module. But really all this feature does is restrict the client to using X25519. A better term might be "NIST", since this feature implies support for P-256 and P-384 only.

cjpatton: Updated.

nox:
Cargo features are supposed to be strictly additive so I'm not sure that's the correct vehicle to use. That being said, I offer no alternative.

cjpatton:
> Cargo features are supposed to be strictly additive so I'm not sure that's the correct vehicle to use.

Are you referring to the fact that "kx-safe-defaul" disables `SslContextBuilder::set_curves()`?

Hmm, I definitely we agree we should try to follow best practice if we can. I guess one option is to put the `set_curves` API behind its own feature. The downside would be yet another feature flag for controlling key exchange preference.

Stepping back, the intent of this PR is to make it easy as possible for our Cloudflare-internal users to enable PQ key exchange without having to manually call `set_curves`. There might be other ways to do this, but based on my own assessment of the code (and my abilities with Rust), this is the lowest-complexity option with the fewest side-effects. See https://github.com/cloudflare/boring/pull/145#discussion_r1295802435.

inikulin:
@nox @cjpatton my understanding is that these features are mutually exclusive, so if we add a compile time check that ensures that features are not enabled simultaneously, won't it solve the conceptual problem?

cjpatton:
> @nox @cjpatton my understanding is that these features are mutually exclusive, so if we add a compile time check that ensures that features are not enabled simultaneously, won't it solve the conceptual problem?

Yes, they're mutually exclusive. What would we want to check at compile time? That if "kx-safe-defaults" is set, then `set_curves()` is missing? How would we do this?

inikulin:
> Yes, they're mutually exclusive. What would we want to check at compile time? That if "kx-safe-defaults" is set, then `set_curves()` is missing? How would we do this?

in `build.rs` we can do something like:
```rust
let enabled_count = [
    cfg!(feature = "kx-client-pq-supported"),
    cfg!(feature = "kx-client-pq-preferred") 
    cfg!(feature = "kx-client-nist-required")
]
.iter()
.filter(|f| f)
.count();

if enabled_count > 1 {
     panic!("...");
}
```

we can do similar checks for feature dependencies. 
Btw, we should also document feature deps: either this features should implicitly enable PQ patch feature or we need to check if it's enabled in build.rs as shown above and error if not

cjpatton:
> in `build.rs` we can do something like:
> 
> ```rust
> let enabled_count = [
>     cfg!(feature = "kx-client-pq-supported"),
>     cfg!(feature = "kx-client-pq-preferred") 
>     cfg!(feature = "kx-client-nist-required")
> ]
> .iter()
> .filter(|f| f)
> .count();
> 
> if enabled_count > 1 {
>      panic!("...");
> }
> ```

But where would we do this? If I understand correctly what you're after, the goal is something like: if `cfg!(feature = "kx-safe-defaults")` and  `SslContextBuilder::set_curves()` is defined, then fail to build. How do we express the second condition?

Also, note that it suffices to check if "kx-safe-defaults" is set because this is implied by any of the "kx-*" features.

> we can do similar checks for feature dependencies. Btw, we should also document feature deps: either this features should implicitly enable PQ patch feature or we need to check if it's enabled in build.rs as shown above and error if not

Got, will do after rebasing this PR.

cjpatton:
Squashed.

cjpatton:
Changed "nist-required" to "fips-required". Eventually X25519 might be FIPS-validated.

inikulin:
@cjpatton 
> But where would we do this?

I think there was a misunderstanding here, I thought that @nox complained that features are mutually exclusive, i.e. `kx-client-pq-preferred` and `kx-client-nist-required` can't be used together. Apparently it's not a case.

> Changed "nist-required" to "fips-required". Eventually X25519 might be FIPS-validated.

not sure how do I feel about that, this seems to be relatively distant future thing and might be confusing considering currently FIPS-validated revision

cjpatton:
> @cjpatton
> 
> > But where would we do this?
> 
> I think there was a misunderstanding here, I thought that @nox complained that features are mutually exclusive, i.e. `kx-client-pq-preferred` and `kx-client-nist-required` can't be used together. Apparently it's not a case.

Ah, OK! In fact, "kx-client-pq-preferred" and "kx-client-nist-required" _can_ be used at the same time. The effect is that P256Kyber768Draft00 is the client's preferred key exchange algorithm.


> > Changed "nist-required" to "fips-required". Eventually X25519 might be FIPS-validated.
> 
> not sure how do I feel about that, this seems to be relatively distant future thing and might be confusing considering currently FIPS-validated revision

No problem, reverted.

cjpatton:
Fixed up commit message.

PR #147, opened 2023-08-11
Fix build with "fips-link-precompiled" feature

cjpatton:
When the "fips-link-precompiled" feature is used, the build script boringSSL (`boring-sys/build.sh`) adds a precompiled `bcm.o` module provided by the user. The module is renamed to `bcm-fips.o` and inserted into `libcrypto.a` just before the `bcm.o` built by the script. The intent is to "shadow" the module so that, for any symbols that are provided by both, the linker picks the implementations provided by `bcm-fips.o`. At the same time, any sybmols in `bcm.o` that are not in `bcm-fips.o` can be used.

This configuration requires special flags in order to tell the linker how to resolve duplicate symbols (RUSTFLAGS="-Clink-args=-Wl,-zmuldefs" is sufficient). However even with these flags thare are certain symbols that don't resolve. In particular `bcm-fips.o` expects `bcm.o` to provide `RAND_need_entropy`.

Rather than attempt to cobble together a working version of this "shadow" build of `libcrypto.a`, we modify the build script so that it "replaces" `bcm.o` with the precompiled module provided by the user. Based on internal conversations, this appears to be sufficient for every use case for these bindings. If the shadow build is required, then the user will need to provide their own version of `libcrypto.a` (This is not supported as of this commit.)

One more change is required in order to build with "fips-link-precompiled". Building because the FFI exports a different API than the bindings expects. To fix this, it is sufficient to change the features so that "fips-link-precompiled" does not imply "fips".


boring-sys/build.rs
```diff
@@ -538,41 +538,46 @@ fn build_boring_from_sources() -> String {
...
+    const BCM_O_NAME: &str = "bcm.c.o";
+    println!("cargo:warning=linking in precompiled {BCM_O_NAME}` module");
 
-    let bcm_o_src_path = env::var("BORING_SSL_PRECOMPILED_BCM_O")
+    let precompiled_bcm_o_src_path = env::var("BORING_SSL_PRECOMPILED_BCM_O")
         .expect("`fips-link-precompiled` requires `BORING_SSL_PRECOMPILED_BCM_O` env variable to be specified");
 
     let libcrypto_path = PathBuf::from(bssl_dir)
-        .join("build/crypto/libcrypto.a")
+        .join("build/libcrypto.a")
```
cjpatton, line 548: I forgot to mention this in the commit mesesage, but this seems to be necessary for the version of boringSSL we're building.

boring-sys/build.rs
```diff
@@ -538,41 +538,46 @@ fn build_boring_from_sources() -> String {
...
         .canonicalize()
         .unwrap()
         .display()
         .to_string();
 
-    let bcm_o_dst_path = PathBuf::from(bssl_dir).join("build/bcm-fips.o");
```
cjpatton, line 553: This is the critical line. The "shadow" build adds gives the precompiled module a distinctive name so that `bcm-fips.o` and `bcm.o` are both included in `libcrypto.a`. In this PR, we give the precompiled module _the same name_ so that it _replaces_ `bcm.o`.

bwesterb:
lgtm

cjpatton:
I've updated test `fips::is_enabled` to check for FIPS mode (`FIPS_mode() != 0`) when the "fips-link-precompiled" flag is set. It turns out that it is not set, as we expected. While debugging this, @bhalleycf noticed that we're not actually modifying `bcm.c.o` in the build script. After fixing this issue (https://github.com/cloudflare/boring/pull/147/commits/4b8b5d344447173a80e524e99f168e97e3912cb6) the duplicate symbol errors are there, and we're missing `RAND_need_entropy`.

cjpatton:
Closing this PR since we know it doesn't fix the "fips-link-precompiled" build on its own.

PR #152, opened 2023-08-15
Provide into_ssl() for ConnectConfiguration

eaufavor:
Port from openssl-rs.

### DIFF
--- a/boring/src/ssl/connector.rs
+++ b/boring/src/ssl/connector.rs
@@ -8,6 +8,7 @@ use crate::ssl::{
     SslOptions, SslRef, SslStream, SslVerifyMode,
 };
 use crate::version;
+use std::net::IpAddr;
 
 const FFDHE_2048: &str = "
 -----BEGIN DH PARAMETERS-----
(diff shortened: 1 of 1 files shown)


boring/src/ssl/connector.rs
```diff
@@ -189,6 +190,21 @@ impl ConnectConfiguration {
...
+    ///
+    /// The domain is used for SNI (if it is not an IP address) and hostname verification if enabled.
+    pub fn into_ssl(mut self, domain: &str) -> Result<Ssl, ErrorStack> {
+        if self.sni && domain.parse::<IpAddr>().is_err() {
+            self.ssl.set_hostname(domain)?;
+        }
+
+        if self.verify_hostname {
+            setup_verify_hostname(&mut self.ssl, domain)?;
+        }
```
nox, line 203: Those checks are different than in `connect` below, they should be the same and `connect` should call `into_ssl`.

eaufavor: Updated. I keep the `&& domain.parse::<IpAddr>().is_err()` which is a fix ported from openssl-rs.

nox (approved):

PR #156, opened 2023-08-16
Re-enable ability to use pre-compiled FIPS

howardjohn:
This check was added in https://github.com/cloudflare/boring/pull/130, and broke us. 

we currently use precompiled FIPS builds, since we don't want our developers to need to have the ancient/obscure toolchains, etc required to build FIPS boringssl.\

https://github.com/cloudflare/boring/pull/130 looks useful, but as far as I understand this only allows using FIPS boringcrypto + latest boringssl, not FIPS boringcrypto+boringssl at the same SHA. The former may be something we want to move towards in the future, but it seems nice to have both options.

I would not be surprised if I am misunderstanding #130, but there isn't much context associated with it.

### DIFF
--- a/boring-sys/build.rs
+++ b/boring-sys/build.rs
@@ -602,7 +602,7 @@ fn check_feature_compatibility() {
     let is_precompiled_native_lib = env::var("BORING_BSSL_PATH").is_ok();
 
     if is_precompiled_native_lib && build_from_sources_required {
-        panic!("precompiled BoringSSL was provided, so FIPS configuration or optional patches can't be applied");
+        println!("cargo:warning=precompiled BoringSSL was provided, so FIPS configuration or optional patches can't be applied");
     }
 }
 


nmittler:
@nox @inikulin mind taking a look?

inikulin:
@howardjohn the change makes sense for me. Could you rebase it on top of recent changes? I think it's also worth emitting a warning compile time message (via `println!("cargo:warning=...");`) for this configuration warning users that they must ensure that provided binary is actually compiled with fips configuration

howardjohn:
> @howardjohn the change makes sense for me. Could you rebase it on top of recent changes? I think it's also worth emitting a warning compile time message (via `println!("cargo:warning=...");`) for this configuration warning users that they must ensure that provided binary is actually compiled with fips configuration

Yep, added both. Thanks

nox:
@howardjohn Could you tell us exactly how you trigger that error? Are you using feature `fips` or `fips-link-precompiled`? Also, which variables are you defining? Are you using `BORING_BSSL_PATH`?

https://github.com/cloudflare/boring/blob/4749c525e4dc2c3c136589067f146689ce2bc2b1/boring-sys/build.rs#L599-L607

`is_precompiled_native_lib` is true because you are using `BORING_BSSL_PATH`, but why is `build_from_sources_required` true?

If it is because you are enabling feature `fips-link-precompiled`, then just don't, as it is only relevant for the `BORING_SSL_PRECOMPILED_BCM_O` builds as introduced by #130, as features `fips` and `fips-link-precompiled` were separated in #160.

If it is because you are enabling features `pq-experimental` or `rpk`, causing `features_with_patches_enabled` to be true, then also enable the feature `no-patches` and you should be set.

howardjohn:
Hmm, it seems like pulling from latest `boring` fixes this without this PR now.

Our setup:

```
BORING_BSSL_PATH = { value = "vendor/boringssl-fips/linux_x86_64", force = true, relative = true }
BORING_BSSL_INCLUDE_PATH = { value = "vendor/boringssl-fips/include/", force = true, relative = true }
```

with `features=fips`

nox:
Cool, closing this then. Could you check if #170 works for you with

```
BORING_BSSL_FIPS_PATH = { value = "vendor/boringssl-fips/linux_x86_64", force = true, relative = true }
BORING_BSSL_FIPS_INCLUDE_PATH = { value = "vendor/boringssl-fips/include/", force = true, relative = true }
```

PR #166, opened 2023-09-18
Enable P-521 with "kx-safe-default"

cjpatton:
While not commonly used, P-521 is a perfectly safe choice of key exchange algorithm.


eaufavor:
Note that BoringSSL by default doesn't advertise P-521. I don't think this change will do any harm. It is just a bit unexpected.

inikulin (approved):

PR #174, opened 2023-10-12
Introduce set_async_get_session_callback

nox:
The signature of the sync callback required a breaking change.


ghedo:
Ideally there should be a test, like we have for the other async callbacks.

ghedo (approved):

PR #180, opened 2023-10-26
hyper and tokio "full" feature for dev builds only

### LINKED ISSUE #179: hyper and tokio are always used with the "full" feature enabled
In d1ee9bfd868a472a04f92e15e011d0fc42db6b88 cargo workspaces were introduced, but that led to always enabling the `full` features for `tokio` and `hyper`. Previously the `full` feature was only used for dev-dependencies.

shahn:
This was an accidental regression in
d1ee9bfd868a472a04f92e15e011d0fc42db6b88 which leads to pulling in the full featureset of tokio and hyper for all dependents of tokio-boring and hyper-boring.

Fixes #179.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,11 +27,11 @@ libc = "0.2"
 hex = "0.4"
 rusty-hook = "^0.11"
 futures = "0.3"
-tokio = { version = "1", features = ["full"] }
+tokio = { version = "1" }
 anyhow = "1"
 antidote = "1.0.0"
 http = "0.2"
-hyper = { version = "0.14", default-features = false, features = ["full"] }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -27,11 +27,11 @@ libc = "0.2"
 hex = "0.4"
 rusty-hook = "^0.11"
 futures = "0.3"
-tokio = { version = "1", features = ["full"] }
+tokio = { version = "1" }
```
nox, line 30:
```suggestion
tokio = "1"
```

nox (approved):

shahn:
Ah yes, thanks for that simplification!

PR #187, opened 2023-11-11
boring-sys: Don't use CMake cross-compilation for macOS->iOS

jrose-signal:
(or macOS->macOS)

Fixes the iOS build.


nox:
Thanks for the PR. Could you explain what it fixes? Maybe show an example of a broken build with the current code?

jrose-signal:
You can see one example [in this log](https://github.com/signalapp/boring/actions/runs/6869273854/job/18681690799?pr=23) after fixing the CI to test cross-compilation more thoroughly (#188). In this case the assembly sources end up getting built for iOS devices rather than iOS simulators, presumably because CMake assumes you'll set *all* of the relevant cross-compilation properties if you're going to use any of them. But for toolchains that natively support multiple targets, this just makes things worse.

(Honestly, I don't think this sort of policy should be happening here at all; it should be in the `cmake` crate. But [that has similar problems](https://github.com/rust-lang/cmake-rs/pull/193), and since the `cmake` crate isn't being actively developed I understand doing things here!)

jrose-signal:
[Citation from the CMake docs themselves](https://cmake.org/cmake/help/latest/variable/CMAKE_CROSSCOMPILING.html):

> Another case to be aware of is that builds targeting Apple platforms other than macOS are handled differently to other cross compiling scenarios. Rather than relying on [`CMAKE_SYSTEM_NAME`](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_NAME.html#variable:CMAKE_SYSTEM_NAME) to select the target platform, Apple device builds use [`CMAKE_OSX_SYSROOT`](https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_SYSROOT.html#variable:CMAKE_OSX_SYSROOT) to select the appropriate SDK, which indirectly determines the target platform. […] Therefore, the use of `CMAKE_CROSSCOMPILING` is not recommended for projects targeting Apple devices.

nox:
Sorry I had other yaks to shave these last few days but I'll review these tomorrow and so far I don't see a reason not to merge those changes.

jrose-signal:
No worries, that was less a "hey, pay attention to me" and more "I found a citation to support this, let me put in the PR for posterity"!

nox:
I fixed the failing clippy lint in a separate PR so you can just rebase yours for CI to go green.

nox (approved):

PR #188, opened 2023-11-14
Update CI to build for all matrix targets

eager-signal:
We noticed that the CI matrix wasn’t always building for the Rust target it intended, leading to surprise breakage (e.g. #187) when doing release builds. This addresses that in a few ways:

1. add `--target` to `cargo test`
2. `check_only` flag
   - skip running tests for non-native architectures and targets; instead do `cargo build … --tests` to confirm cross-compilation
3. various customization options
   - specific packages required to cross-compile for the target, environment variables, etc.  
4. last resort: drop a target
   - musl doesn’t have a standard C++ setup to use


nox:
Thanks for the impressive PR! Can you please rebase on top of master now that #187 is merged? Thanks.

jrose-signal:
Fixes #62 :-)

nox (approved):

PR #107, opened 2023-03-10
hyper-boring: update to hyper 1.0.0

howardjohn:
This PR upgrades hyper-boring to support hyper 1.0.0-rc.3. Most of this is by moving dependencies from hyper core to hyper-util, as much of the functionality we depend on has been removed from the core.

PR #194, opened 2023-11-27
hyper-boring: update to hyper 1.0.0

howardjohn:
This PR upgrades hyper-boring to support hyper 1.0.0-rc.3. Most of this is by moving dependencies from hyper core to hyper-util, as much of the functionality we depend on has been removed from the core.

Copied from https://github.com/cloudflare/boring/pull/107 since I made an unrepairable mistake on the first PR

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
-version = "4.1.0"
+version = "5.0.0"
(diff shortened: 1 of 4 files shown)


howardjohn:
@inikulin we have moved off using hyper-boring so its not a high priority for me; if anyone is interested feel free to take over this PR/make a new one based on this. If I have some free time I may be able to pick it up, but don't want to block everyone on that.

PR #212, opened 2024-01-16
Expose `set_compliance_policy` and `get_ciphers`

jhoyla:
Expose APIs to access `SSL_CTX_set_compliance_policy` and `SSL_CTX_get_ciphers`.

### DIFF
--- a/boring/src/ssl/mod.rs
+++ b/boring/src/ssl/mod.rs
@@ -89,7 +89,7 @@ use crate::srtp::{SrtpProtectionProfile, SrtpProtectionProfileRef};
 use crate::ssl::bio::BioMethod;
 use crate::ssl::callbacks::*;
 use crate::ssl::error::InnerError;
-use crate::stack::{Stack, StackRef};
+use crate::stack::{Stack, StackRef, Stackable};
 use crate::x509::store::{X509Store, X509StoreBuilderRef, X509StoreRef};
 use crate::x509::verify::X509VerifyParamRef;
 use crate::x509::{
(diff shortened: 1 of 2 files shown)


boring/src/ssl/mod.rs
```diff
@@ -701,6 +701,27 @@ impl SslCurve {
     pub const P256_KYBER768_DRAFT00: SslCurve = SslCurve(ffi::NID_P256Kyber768Draft00);
 }
 
+/// A compliance policy
```
ghedo, line 704:
```suggestion
/// A compliance policy.
```

boring/src/ssl/mod.rs
```diff
@@ -1281,6 +1304,21 @@ impl SslContextBuilder {
...
 
+    /// Gets the list of supported ciphers for protocols before TLSv1.3.
+    ///
+    /// See [`ciphers`] for details on the format
+    ///
+    /// This corresponds to [`SSL_CTX_get_ciphers`].
+    ///
+    /// [`ciphers`]: https://www.openssl.org/docs/manmaster/man1/ciphers.html
+    /// [`SSL_CTX_set_cipher_list`]: https://www.openssl.org/docs/manmaster/man3/SSL_CTX_get_ciphers.html
+    pub fn get_ciphers(&mut self) -> Result<&StackRef<SslCipher>, ErrorStack> {
```
nox, line 1315: This should take `&self` and you may want to put the same method on `SslContext` too.

nox, line 1315: I don't think the Result is necessary here

https://github.com/google/boringssl/blob/b628f8721e7e67302b0e26c92e0108a31066194e/ssl/ssl_lib.cc#L2075-L2077

Seems like NULL should be interpreted as None, or even maybe that the method should panic on NULL.

https://github.com/google/boringssl/blob/master/ssl/internal.h#L631

nox (approved):

boring/src/ssl/mod.rs
```diff
@@ -1281,6 +1304,25 @@ impl SslContextBuilder {
...
 
+    /// Gets the list of supported ciphers for protocols before TLSv1.3.
+    ///
+    /// See [`ciphers`] for details on the format
+    ///
+    /// This corresponds to [`SSL_CTX_get_ciphers`].
+    ///
+    /// [`ciphers`]: https://www.openssl.org/docs/manmaster/man1/ciphers.html
+    /// [`SSL_CTX_set_cipher_list`]: https://www.openssl.org/docs/manmaster/man3/SSL_CTX_get_ciphers.html
+    pub fn ciphers(&self) -> Option<&StackRef<SslCipher>> {
```
nox, line 1315: `SslContextBuilder` owns a `SslContext`, so this should just be `self.0.ciphers()`

nox (changes requested):

nox (approved):

PR #215, opened 2024-01-24
Fix building with non bazel commits of boringssl

rushilmehra:
We need to add `/build/crypto` and `/build/ssl` to the library search path to handle the case where we pass `BORING_BSSL_SOURCE_PATH` when building without enabling any fips features. Otherwise, non bazel commits will not work because `/build/` itself will not contain any crypto libraries to link with

### DIFF
--- a/boring-sys/build/config.rs
+++ b/boring-sys/build/config.rs
@@ -5,6 +5,7 @@ use std::path::PathBuf;
 pub(crate) struct Config {
     pub(crate) manifest_dir: PathBuf,
     pub(crate) out_dir: PathBuf,
+    pub(crate) is_bazel: bool,
     pub(crate) host: String,
     pub(crate) target: String,
     pub(crate) target_arch: String,
@@ -51,9 +52,15 @@ impl Config {
             features.fips || features.fips_link_precompiled,
         );
 
+        let mut is_bazel = false;
+        if let Some(src_path) = &env.source_path {
+            is_bazel = src_path.join("src").exists();
+        }
+
         let config = Self {
             manifest_dir,
             out_dir,
+            is_bazel,
             host,
             target,
             target_arch,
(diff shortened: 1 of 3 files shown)


nox:
> We need to add `/build/crypto` and `/build/ssl` to the library search path to handle the case where we pass `BORING_BSSL_SOURCE_PATH` when building without enabling any fips features.

There are other issues when you do that, patches won't apply for example. The fix needs to be a bit more complex than that, e.g. require ASSUME_PATCHED if features that require patches are enabled.

I'd also rather tweak the build to check if it is a bazel build or not, and adjust the paths accordingly.

rushilmehra:
> > We need to add `/build/crypto` and `/build/ssl` to the library search path to handle the case where we pass `BORING_BSSL_SOURCE_PATH` when building without enabling any fips features.
> 
> There are other issues when you do that, patches won't apply for example. The fix needs to be a bit more complex than that, e.g. require ASSUME_PATCHED if features that require patches are enabled.
> 
> I'd also rather tweak the build to check if it is a bazel build or not, and adjust the paths accordingly.

 I updated the PR to require BORING_BSSL_ASSUME_PATCHED if you pass BORING_BSSL_SOURCE_PATH and depend on a feature that would apply a patch. As for updating paths based on the boringssl passed being bazel or not, I'm going to figure that out in a follow up PR because it's going to be a bit annoying. All the git patches assume a src/ directory is available, but for non bazel builds that isn't the case. I have a feeling I'll have to redo the patches to remove the src/ and just set src/ as the working directory for bazel versions of boringssl

rushilmehra:
Okay now we pass -p2 to git apply when we identify a non bazel build. I tested this by changing the boringssl submodule to both a bazel and bazel build, and enabling the rpk feature.

As an extra goodie, I filed https://github.com/cloudflare/boring/pull/217 to make builds faster

ghedo (approved):

PR #216, opened 2024-01-25
Removes vestigial build script

ibeckermayer:
`boring-sys`'s build script is pointed at `"build/main.rs"`, which afaict means that `boring-sys/build.rs` is leftover and no longer in use. (This caused a lot of confusion for me in modifying the build script for Teleport's fork of the project).

https://github.com/cloudflare/boring/blob/master/boring-sys/Cargo.toml#L12


rushilmehra (approved):
Yeah I spent a few hours the other day putting debug logs in the wrong build script :) lgtm but will defer to @nox in case there's a reason it's still there

ibeckermayer:
Friendly ping @nox

PR #217, opened 2024-01-27
Set CMAKE_BUILD_PARALLEL_LEVEL to num_cpus

rushilmehra:
cmake-rs' jobserver doesn't work reliably, if at all. One workaround is to set CMAKE_BUILD_PARALLEL_LEVEL to num_cpus. On my machine it makes builds over 3x faster


rushilmehra:
Happy to hide this behind a feature flag if deemed necessary

kornelski (approved):

kornelski:
num_cpus is obsolete now: https://doc.rust-lang.org/stable/std/thread/fn.available_parallelism.html

PR #223, opened 2024-02-06
fix: invalid -L path for build artifact search

poliorcetics:
I haven't managed to link boring for either Linux, windows or macOS without this modification recently, which I find strange

Do you have any idea of what could be happening ?

### DIFF
--- a/boring-sys/build/main.rs
+++ b/boring-sys/build/main.rs
@@ -645,19 +645,8 @@ fn main() {
         );
     } else {
         println!(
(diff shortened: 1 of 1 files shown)


rushilmehra:
What's the build error? And what version of boringssl are you building with?

Guiguiprim:
The error is `error: could not find native static library `crypto`, perhaps an -L flag is missing?` when building an app using boring 4.5.0 on windows.

PaulDance:
Just tested it under 4.12.0 from crates.io and the issue persists.

PR #227, opened 2024-03-21
Add getters for client hello raw data

jrouviere:


rushilmehra (approved):

rushilmehra:
lgtm, @nox want to take a quick look?

boring/src/ssl/mod.rs
```diff
@@ -2240,6 +2240,16 @@ impl ClientHello<'_> {
     pub fn version_str(&self) -> &'static str {
         self.ssl().version_str()
     }
+
+    /// Returns the client hello raw data
+    pub fn raw_client_hello(&self) -> &[u8] {
```
nox, line 2245: This is the entire client hello record, right? Then I would just name that as `as_bytes`.

boring/src/ssl/mod.rs
```diff
@@ -2240,6 +2240,16 @@ impl ClientHello<'_> {
...
+        unsafe { slice::from_raw_parts(self.0.client_hello, self.0.client_hello_len) }
+    }
+
+    /// Returns the client random raw data
+    pub fn raw_random(&self) -> &[u8] {
```
nox, line 2250: I think `random` is a better name, "raw" feels weird to me as there could not be a non-raw version of this, right?

boring/src/ssl/mod.rs
jrouviere: Yes that's the whole message sent by the client. Agreed that's a better name, will change.

boring/src/ssl/mod.rs
jrouviere: Good point, not much you can parse, I'll rename it.

nox (approved):

PR #229, opened 2024-03-24
Remove kx-safe-default gate on SslCurve

rushilmehra:
While setting curves should be restricted by the kx-safe-default feature, reading the curve is allowed.


inikulin (approved):

bwesterb (approved):

cjpatton (approved):
Exposing the SSL curves list is fine, but we still want to prevent the user from setting the curves list if the "kx-*" feature flags are in use. This is because the features will silently overwrite the user's preference.

PR #239, opened 2024-06-03
Fix SSL curve identifiers

jrouviere:
While testing I noticed that `SslRef::curve()` was returning `29` (SSL_CURVE_X25519) instead of `948` (NID_X25519).
I assume most users rely on `SslCurve::name()` which is correct because it's passing the id back to the ffi layer.


bwesterb (approved):

bwesterb:
This breaks [`set_curves`](https://github.com/cloudflare/boring/blob/b3eaacc33a040597e1b2bf45008927f8a368ac6e/boring/src/ssl/mod.rs#L1896).

bwesterb (approved):

PR #240, opened 2024-06-14
Expose various APIs necessary for certificate selection logic

rushilmehra:


boring/src/ssl/mod.rs
```diff
@@ -2264,11 +2268,29 @@ impl ClientHello<'_> {
     pub fn random(&self) -> &[u8] {
         unsafe { slice::from_raw_parts(self.0.random, self.0.random_len) }
     }
+
+    /// Returns the raw list of ciphers supported by the client in its Client Hello record.
+    pub fn ciphers(&self) -> &[u8] {
```
ghedo, line 2273: This just returns the raw bytes, but really ciphers are `u16`, so it would be more helpful if it just returned a slice of `u16` instead, but I'm not entirely sure _how_ to do that without re-allocating the whole thing though :laughing: ... there's probably some kind of unsavoury trick to do it in Rust...

boring/src/ssl/mod.rs
```diff
@@ -3109,6 +3166,78 @@ impl SslRef {
         str::from_utf8(version.to_bytes()).unwrap()
     }
 
+    /// Sets the minimum supported protocol version.
+    ///
+    /// A value of `None` will enable protocol versions down the the lowest version supported by
+    /// OpenSSL.
```
ghedo, line 3172: :see_no_evil: 

We might want to stop copying docs from OpenSSL...

rushilmehra: Fixed, after investigation it seems that BoringSSL won't ever return `0` for `get_{min|max|_proto_version`, so those APIs returning Options doesn't make much sense. We should fix that in the next breaking change

boring/src/ssl/mod.rs
rushilmehra: So unfortunately casting the raw `*const u8` to a `*const u16` doesn't work, because of pointer alignment issues. It's problematic because the underlying `*const u8` comes form the middle of another buffer, specifically the one that holds the incoming Client Hello message body. I think we'll have to copy these bytes into another u16 buffer - do we want to do that? Or do we want to pass the u8 slice to the user and let them parse it themselves?

ghedo (approved):

PR #241, opened 2024-06-18
Expose API to enable certificate compression.

mstyura:
As a user of `tokio-boring` I'd like to control certificate compression feature using high-level Rust-friendly API instead of directly calling C functions form `boring-sys`.
I'm ready to adjust implementation if necessary to match the style of the rest of the crate.


mstyura:
Hello @inikulin!
Could you please review once you have time?
Thanks a lot in advance!

boring/src/ssl/callbacks.rs
```diff
@@ -579,3 +579,94 @@ pub(super) unsafe extern "C" fn raw_info_callback<F>(
...
+    let input_slice = unsafe { std::slice::from_raw_parts(input, input_len) };
+
+    let output_slice = unsafe { std::slice::from_raw_parts_mut(data, uncompressed_len) };
+
+    let mut cursor = std::io::Cursor::new(output_slice);
+    if compressor.decompress(input_slice, &mut cursor).is_err() {
+        return 0;
+    }
+    if cursor.position() != uncompressed_len as u64 {
+        return 0;
```
kornelski, line 647: This leaks `decompressed`?

boring/src/ssl/mod.rs
```diff
@@ -1594,6 +1604,46 @@ impl SslContextBuilder {
...
+        C: CertificateCompressor,
+    {
+        let algo = compressor.algorithm();
+        let (can_compress, can_decompress) =
+            (compressor.can_compress(), compressor.can_decompress());
+        assert!(can_compress || can_decompress);
+        let success = unsafe {
+            self.replace_ex_data(SslContext::cached_ex_index::<C>(), compressor);
+
+            ffi::SSL_CTX_add_cert_compression_alg(
```
kornelski, line 1626: The callbacks are per algorithm, but the ex data is per type. If `can_compress`/`can_decompress` were set dynamically, it could get the callbacks out of sync with the implementation.

Maybe this function should fail if the ex data for `C` already exists?

Also `SSL_CTX_add_cert_compression_alg` will fail if the same algorithm is set more than once, but keeps the previous callbacks still set. In the current implementation that would use old callbacks, but replace the instance of `C` regardless. It's not strictly speaking unsound since the callbacks are type-specific, but it may be surprising.

kornelski (changes requested):
Waiting for response from the author

boring/src/ssl/callbacks.rs
mstyura: Thanks for noticing, I've introduced wrapper type which I hope ensures correct usage of `CRYPTO_BYFFER` instance.

boring/src/ssl/mod.rs
yyaroshevich: > The callbacks are per algorithm, but the ex data is per type. If can_compress/can_decompress were set dynamically, it could get the callbacks out of sync with the implementation.

I've changed algorithm, can_compress and can_decompress to be the associated constants of trait.

> Maybe this function should fail if the ex data for C already exists?

I've changed the order of operations to only override ex data if `SSL_CTX_add_cert_compression_alg` succeed.

> Also SSL_CTX_add_cert_compression_alg will fail if the same algorithm is set more than once, but keeps the previous callbacks still set. In the current implementation that would use old callbacks, but replace the instance of C regardless. It's not strictly speaking unsound since the callbacks are type-specific, but it may be surprising.

Could you please check that with the latest changes this is no longer an issue?

mstyura: > The callbacks are per algorithm, but the ex data is per type. If can_compress/can_decompress were set dynamically, it could get the callbacks out of sync with the implementation.

I've changed algorithm, can_compress and can_decompress to be the associated constants of trait.

> Maybe this function should fail if the ex data for C already exists?

I've changed the order of operations to only override ex data if `SSL_CTX_add_cert_compression_alg` succeed.

> Also SSL_CTX_add_cert_compression_alg will fail if the same algorithm is set more than once, but keeps the previous callbacks still set. In the current implementation that would use old callbacks, but replace the instance of C regardless. It's not strictly speaking unsound since the callbacks are type-specific, but it may be surprising.

Could you please check that with the latest changes this is no longer an issue?

boring/src/ssl/callbacks.rs
```diff
@@ -579,3 +579,144 @@ pub(super) unsafe extern "C" fn raw_info_callback<F>(
...
+            std::mem::forget(self);
+            Ok(result)
+        }
+    }
+}
+
+impl<'a> Drop for CryptoBufferBuilder<'a> {
+    fn drop(&mut self) {
+        unsafe {
+            boring_sys::CRYPTO_BUFFER_free(self.buffer);
```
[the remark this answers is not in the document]
mstyura: Sure, I've added check.

boring/src/ssl/callbacks.rs
```diff
@@ -579,3 +579,146 @@ pub(super) unsafe extern "C" fn raw_info_callback<F>(
...
+
+struct CryptoBufferBuilder<'a> {
+    buffer: *mut ffi::CRYPTO_BUFFER,
+    cursor: std::io::Cursor<&'a mut [u8]>,
+}
+
+impl<'a> CryptoBufferBuilder<'a> {
+    fn with_capacity(capacity: usize) -> Result<CryptoBufferBuilder<'a>, ErrorStack> {
+        let mut data: *mut u8 = std::ptr::null_mut();
+        let buffer = unsafe { crate::cvt_p(ffi::CRYPTO_BUFFER_alloc(&mut data, capacity))? };
```
kornelski, line 690: This data isn't initialized, so theoretically it shouldn't be used with a Rust slice. However, the alternatives would require reinventing `Cursor`.

kornelski (approved):

PR #243, opened 2024-06-24
PQ: fix timing sidechannels and add IPDWing

bwesterb:
Fix three potential timing sidechannels. These don't affect ephemeral usage of Kyber as in TLS, but it's good practice to get rid of them anyway.

Also adds IPDWing, a preliminary version of X-Wing using the initial public draft (IPD) of ML-KEM. Don't use it.

### DIFF
--- a/boring-sys/patches/boring-pq.patch
+++ b/boring-sys/patches/boring-pq.patch
@@ -1,4 +1,4 @@
-From 4cba2164726c8d2647e38548a266a70c4942d567 Mon Sep 17 00:00:00 2001
+From 836d390deaf8b50fed0cafd55b17a63e80454d7f Mon Sep 17 00:00:00 2001
 From: Bas Westerbaan <bas@cloudflare.com>
 Date: Fri, 22 Jul 2022 16:43:48 +0200
 Subject: [PATCH] Add temporary post-quantum key agreements
@@ -20,43 +20,55 @@ This patch adds:
    key agreement should only be used for testing: to see if the smaller
    keyshare makes a difference.
 
+4. Supportfor IPDWing under codepoint 0xfe41. This key agreement
+   is a preliminary version of X-Wing using the initial public draft
+   of ML-KEM. It should not be used.
+
 The patch also replaces Google's implementation of Kyber, by the
 portable reference implementation, so as to support Kyber512.
 
 Cf RTG-2076 RTG-2051 RTG-2508 RTG-2707 RTG-2607
 ---
- BUILD.generated.bzl               |    5 +-
+ BUILD.generated.bzl               |    7 +-
  BUILD.generated_tests.bzl         |    4 -
- CMakeLists.txt                    |    4 +-
- sources.json                      |    9 +-
- src/crypto/CMakeLists.txt         |    5 +-
+ CMakeLists.txt                    |    6 +-
+ sources.json                      |   11 +-
+ src/crypto/CMakeLists.txt         |    7 +-
+ src/crypto/kyber/fips202.c        |  699 +++++++
+ src/crypto/kyber/fips202.h        |   29 +
  src/crypto/kyber/internal.h       |   91 -
+ src/crypto/kyber/ipdwing.c        |  110 ++
  src/crypto/kyber/keccak.c         |  204 --
- src/crypto/kyber/kyber.c          | 2865 ++++++++++++++++++++-------
+ src/crypto/kyber/kyber.c          | 2319 +++++++++++++++-------
+ src/crypto/kyber/kyber.h          |   29 +
  src/crypto/kyber/kyber512.c       |    5 +
  src/crypto/kyber/kyber768.c       |    4 +
  src/crypto/kyber/kyber_test.cc    |  229 ---
- src/crypto/obj/obj_dat.h          |   14 +-
- src/crypto/obj/obj_mac.num        |    3 +
- src/crypto/obj/objects.txt        |    5 +-
- src/include/openssl/kyber.h       |  199 +-
- src/include/openssl/nid.h         |    9 +
- src/include/openssl/ssl.h         |    3 +
+ src/crypto/obj/obj_dat.h          |   17 +-
+ src/crypto/obj/obj_mac.num        |    4 +
+ src/crypto/obj/objects.txt        |    6 +-
+ src/include/openssl/kyber.h       |  252 ++-
+ src/include/openssl/nid.h         |   12 +
+ src/include/openssl/ssl.h         |    4 +
  src/sources.cmake                 |    2 -
- src/ssl/extensions.cc             |    3 +
- src/ssl/ssl_key_share.cc          |  412 +++-
+ src/ssl/extensions.cc             |    4 +
+ src/ssl/ssl_key_share.cc          |  493 ++++-
  src/ssl/ssl_lib.cc                |    2 +-
- src/ssl/ssl_test.cc               |   25 +-
+ src/ssl/ssl_test.cc               |   29 +-
  src/tool/speed.cc                 |  162 +-
- 26 files changed, 2797 insertions(+), 5447 deletions(-)
+ 30 files changed, 3276 insertions(+), 5445 deletions(-)
+ create mode 100644 src/crypto/kyber/fips202.c
+ create mode 100644 src/crypto/kyber/fips202.h
  delete mode 100644 src/crypto/kyber/internal.h
+ create mode 100644 src/crypto/kyber/ipdwing.c
  delete mode 100644 src/crypto/kyber/keccak.c
+ create mode 100644 src/crypto/kyber/kyber.h
  create mode 100644 src/crypto/kyber/kyber512.c
  create mode 100644 src/crypto/kyber/kyber768.c
  delete mode 100644 src/crypto/kyber/kyber_test.cc
 
 diff --git a/BUILD.generated.bzl b/BUILD.generated.bzl
-index 738e1055f..9466757a2 100644
+index 738e1055f..d1d232399 100644
 --- a/BUILD.generated.bzl
 +++ b/BUILD.generated.bzl
 @@ -253,7 +253,6 @@ crypto_internal_headers = [
@@ -67,14 +79,16 @@ index 738e1055f..9466757a2 100644
      "src/crypto/lhash/internal.h",
      "src/crypto/obj/obj_dat.h",
      "src/crypto/pkcs7/internal.h",
-@@ -382,8 +381,8 @@ crypto_sources = [
+@@ -382,8 +381,10 @@ crypto_sources = [
      "src/crypto/fipsmodule/fips_shared_support.c",
      "src/crypto/hpke/hpke.c",
      "src/crypto/hrss/hrss.c",
 -    "src/crypto/kyber/keccak.c",
 -    "src/crypto/kyber/kyber.c",
++    "src/crypto/kyber/fips202.c",
 +    "src/crypto/kyber/kyber512.c",
 +    "src/crypto/kyber/kyber768.c",
++    "src/crypto/kyber/ipdwing.c",
      "src/crypto/lhash/lhash.c",
      "src/crypto/mem.c",
      "src/crypto/obj/obj.c",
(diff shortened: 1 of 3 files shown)


boring-sys/patches/boring-pq.patch
```diff
@@ -4558,10 +4994,89 @@ index 09a9ad380..f7d2226e3 100644
...
++};
++
++class IPDWingKeyShare : public SSLKeyShare {
++ public:
++  IPDWingKeyShare() {}
++
++  uint16_t GroupID() const override { return SSL_CURVE_IPDWING; }
++
++  bool Generate(CBB *out) override {
++    uint8_t entropy[IPDWING_GENERATE_KEY_BYTES];
```
cjpatton, line 5006: "seed" is better aligned with with the draft.
```suggestion
+    uint8_t entropy[IPDWING_GENERATE_SEED_BYTES];
```

boring-sys/patches/boring-pq.patch
```diff
@@ -4558,10 +4994,89 @@ index 09a9ad380..f7d2226e3 100644
...
++    }
++
++    return true;
++  }
++
++  bool Encap(CBB *out_public_key, Array<uint8_t> *out_secret,
++              uint8_t *out_alert, Span<const uint8_t> peer_key) override {
++    Array<uint8_t> secret;
++    *out_alert = SSL_AD_INTERNAL_ERROR;
++    if (!secret.Init(IPDWING_KEY_BYTES)) {
```
cjpatton, line 5025: As above
```suggestion
+    Array<uint8_t> shared_secret;
+    *out_alert = SSL_AD_INTERNAL_ERROR;
+    if (!secret.Init(IPDWING_SHARED_SECRET_BYTES)) {
```

boring-sys/patches/boring-pq.patch
```diff
@@ -3957,18 +4343,39 @@ index cafae9d17..074ac5906 100644
...
--
++struct KYBER512_public_key {
++  uint8_t opaque[KYBER512_PUBLIC_KEY_BYTES];
++};
++struct KYBER768_public_key {
++  uint8_t opaque[KYBER768_PUBLIC_KEY_BYTES];
++};
++struct IPDWING_public_key {
++    struct KYBER768_public_key m;
++    uint8_t x[32];
```
cjpatton, line 4354: nit: I find it odd that the curve point is named `x` here but named `xpub` as part of the private key. In the private key, `x` is the name of the secret scalar. Could we go with `xpub` here instead so that we have one name for each thing?

boring-sys/patches/boring-pq.patch
```diff
@@ -3981,6 +4388,12 @@ index cafae9d17..074ac5906 100644
 +    struct KYBER768_public_key *out_pub, struct KYBER768_private_key *out_priv,
 +    const uint8_t input[KYBER_GENERATE_KEY_BYTES]);
 +
++// IPDWING_generate_key is a deterministic function that outputs a public and
++// private key based on the given entropy.
```
cjpatton, line 4392: Following boringSSL comment conventions, I think you'd want to say here what values are written to which variables.

boring-sys/patches/boring-pq.patch
```diff
@@ -3997,6 +4410,14 @@ index cafae9d17..074ac5906 100644
 +                              const struct KYBER768_public_key *in_pub,
 +                              const uint8_t in[KYBER_ENCAP_BYTES]);
 +
++// IPDWING_encap is a deterministic function the generates and encrypts a random
++// session key from the given entropy, writing those values to |out_shared_key|
++// and |out_ciphertext|, respectively.
++OPENSSL_EXPORT void IPDWING_encap(uint8_t out_ciphertext[IPDWING_CIPHERTEXT_BYTES],
++                              uint8_t out_shared_key[IPDWING_KEY_BYTES],
```
cjpatton, line 4417: Unless boringSSL's variable naming conventions dictate otehrwise, I'd go with "shared_secret" here in order to align with the draft.

boring-sys/patches/boring-pq.patch
```diff
@@ -4025,21 +4456,29 @@ index cafae9d17..074ac5906 100644
...
++
 +// KYBER512_parse_public_key sets |*out| to the public-key encoded in |in|.
 +OPENSSL_EXPORT void KYBER512_parse_public_key(
 +    struct KYBER512_public_key *out, const uint8_t in[KYBER512_PUBLIC_KEY_BYTES]);
 +
 +// KYBER768_parse_public_key sets |*out| to the public-key encoded in |in|.
 +OPENSSL_EXPORT void KYBER768_parse_public_key(
 +    struct KYBER768_public_key *out, const uint8_t in[KYBER768_PUBLIC_KEY_BYTES]);
++
++// IPDWING_parse_public_key sets |*out| to the public-key encoded in |in|.
```
cjpatton, line 4471: Why do we have a parser function for the public key but not the ciphertext?

boring-sys/patches/boring-pq.patch
```diff
@@ -154,3178 +172,3184 @@ index 4c0048e1d..f6ea5c40f 100644
...
+diff --git a/src/crypto/kyber/fips202.c b/src/crypto/kyber/fips202.c
+new file mode 100644
+index 000000000..9713a4f7e
+--- /dev/null
++++ b/src/crypto/kyber/fips202.c
+@@ -0,0 +1,699 @@
++/* Based on the public domain implementation in crypto_hash/keccakc512/simple/ from
++ * http://bench.cr.yp.to/supercop.html by Ronny Van Keer and the public domain "TweetFips202"
++ * implementation from https://twitter.com/tweetfips202 by Gilles Van Assche, Daniel J. Bernstein,
++ * and Peter Schwabe */
```
cjpatton, line 216: Does this file need a copyright notice? Likewise for the header and ipdwing.c.

cjpatton (commented):
I'm working my way down from the TLS bits. Everything looks good so far. My next step is to look at what's been modified in fips202.c, then ipdwing.c itself. Shouldn't be too bad.

A couple of questions:

1. Can you provide a source for the side channel mitigations? I want to make sure I understand what they're doing and that they've been implemented properly.
2. I think it would be worth pulling in the test vectors from the xwing draft. Would you mind adding these to the patch, as well as code to exercise them in tests?

boring-sys/patches/boring-pq.patch
bwesterb: But not with the existing patch.

boring-sys/patches/boring-pq.patch
bwesterb: Because the public key can hide precomputed values like the matrix A. (The patch doesn't do that though.)

boring-sys/patches/boring-pq.patch
bwesterb: Implied by LICENSE-MIT file two folders up.

boring-sys/patches/boring-pq.patch
```diff
@@ -20,43 +20,55 @@ This patch adds:
...
+ CMakeLists.txt                    |    6 +-
+ sources.json                      |   11 +-
+ src/crypto/CMakeLists.txt         |    7 +-
+ src/crypto/kyber/fips202.c        |  699 +++++++
+ src/crypto/kyber/fips202.h        |   29 +
  src/crypto/kyber/internal.h       |   91 -
+ src/crypto/kyber/ipdwing.c        |  110 ++
  src/crypto/kyber/keccak.c         |  204 --
- src/crypto/kyber/kyber.c          | 2865 ++++++++++++++++++++-------
+ src/crypto/kyber/kyber.c          | 2319 +++++++++++++++-------
```
cjpatton, line 42: Reviewer note: Most of the changes here are moving SHA-3 out of this file into fips202.c. Also, some functions have been added for SHAKE256 (needed for IPD); and some unused functions have been pruned.

cjpatton (commented):
I've reviewed up to IPDWing and everything looks good. Again, I would strongly reocommend pulling in the test vectors from the draft.

What remains to check:
1. It looks like this there is a new abosrb-then-squeeze API for SHAKE25.
2. The side channel mitigations. Can you please provide a reference for this? I don't really know what I'm looking at. An in-line comment with a pointer would be most helpful.

boring-sys/patches/boring-pq.patch
```diff
@@ -3391,11 +3486,13 @@ index 776c085f9..346d4daec 100644
 +  memcpy(sk+KYBER_SECRETKEYBYTES-KYBER_SYMBYTES, seed+KYBER_SYMBYTES, KYBER_SYMBYTES);
 +}
 +
-+// Modified crypto_kem_enc to BoringSSL style API
-+void encap(uint8_t out_ciphertext[KYBER_CIPHERTEXTBYTES],
++// Internal version that allows us to select between initial public draft
++// (when ipd=1) and round3 kyber (ipd=0).
++void encap2(uint8_t out_ciphertext[KYBER_CIPHERTEXTBYTES],
```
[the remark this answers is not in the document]
bwesterb: Yes.

bwesterb:
> I've reviewed up to IPDWing and everything looks good. Again, I would strongly reocommend pulling in the test vectors from the draft.
> 
> What remains to check:
> 
> 1. It looks like this there is a new abosrb-then-squeeze API for SHAKE25.

It's re-introduced. Upstream [always contained the functions](https://github.com/pq-crystals/kyber/blob/main/ref/fips202.c#L637), but didn't use it, so I pruned it for this patch.

> 2. The side channel mitigations. Can you please provide a reference for this? I don't really know what I'm looking at. An in-line comment with a pointer would be most helpful.

For 2/3 division is replaced by Barrett reduction, which is the exact same fix as upstream ([1](https://github.com/pq-crystals/kyber/commit/dda29cc63af721981ee2c831cf00822e69be3220), [2](https://github.com/pq-crystals/kyber/commit/272125f6acc8e8b6850fd68ceb901a660ff48196)).

For the third, we use BoringSSL's internal value barrier to prevent the compiler from using the fact that an intermediate is {0,1}-valued. Upstream fixes that in a [less robust](https://github.com/pq-crystals/kyber/commit/9b8d30698a3e7449aeb34e62339d4176f11e3c6c) way that presumes that one does not use LTO.

boring-sys/patches/boring-pq.patch
```diff
@@ -154,3178 +172,3184 @@ index 4c0048e1d..f6ea5c40f 100644
...
++        state[23] = Aso;
++        state[24] = Asu;
++}
 +
-+static void polyvec_compress(uint8_t r[KYBER_POLYVECCOMPRESSEDBYTES], const polyvec *a);
-+static void polyvec_decompress(polyvec *r, const uint8_t a[KYBER_POLYVECCOMPRESSEDBYTES]);
 +
-+static void polyvec_tobytes(uint8_t r[KYBER_POLYVECBYTES], const polyvec *a);
-+static void polyvec_frombytes(polyvec *r, const uint8_t a[KYBER_POLYVECBYTES]);
++/*************************************************
```
cjpatton, line 557: This PR adds keccak_init, keccak_absorb, keccak_finalize, and keccak_squeeze from https://github.com/pq-crystals/kyber/blob/main/ref/fips202.c#L637. The code is the same.

nit: keccak_squeeze now appears out of order in this file: in the source it follows keccak_finalize.

Likewise for shake256 below.

boring-sys/patches/boring-pq.patch
```diff
@@ -154,3178 +172,3184 @@ index 4c0048e1d..f6ea5c40f 100644
...
++*              - const poly *a: pointer to input polynomial
 +**************************************************/
-+static void pack_sk(uint8_t r[KYBER_INDCPA_SECRETKEYBYTES], polyvec *sk)
++static void poly_compress(uint8_t r[KYBER_POLYCOMPRESSEDBYTES], const poly *a)
 +{
-+  polyvec_tobytes(r, sk);
-+}
++  unsigned int i,j;
++  int16_t u;
++  uint32_t d0;
```
cjpatton, line 2177: You use `uint64_t` for the same variable below in `polyvec_compress()`. Is `uint32_t` large enough here?

cjpatton (approved):
Everything looks correct to me. Readability could be improved (see inline comments), but if you want you can ignore those.

> > I've reviewed up to IPDWing and everything looks good. Again, I would strongly reocommend pulling in the test vectors from the draft.
> 
> I cross tested connections with the [Go implementation](https://github.com/cloudflare/go/tree/bas/ipdwing), which tests against the test vectors.

Fair enough, but implementations can drift. In my opinion, you should either add that test to CI here or add consume the test vectors here as well.

> > What remains to check:
> > 
> > 1. It looks like this there is a new abosrb-then-squeeze API for SHAKE25.
> 
> It's re-introduced. Upstream [always contained the functions](https://github.com/pq-crystals/kyber/blob/main/ref/fips202.c#L637), but didn't use it, so I pruned it for this patch.

Thanks! I confirmed the code is the same.

> > 2. The side channel mitigations. Can you please provide a reference for this? I don't really know what I'm looking at. An in-line comment with a pointer would be most helpful.
> 
> For 2/3 division is replaced by Barrett reduction, which is the exact same fix as upstream ([1](https://github.com/pq-crystals/kyber/commit/dda29cc63af721981ee2c831cf00822e69be3220), [2](https://github.com/pq-crystals/kyber/commit/272125f6acc8e8b6850fd68ceb901a660ff48196)).
> 
> For the third, we use BoringSSL's internal value barrier to prevent the compiler from using the fact that an intermediate is {0,1}-valued. Upstream fixes that in a [less robust](https://github.com/pq-crystals/kyber/commit/9b8d30698a3e7449aeb34e62339d4176f11e3c6c) way that presumes that one does not use LTO.

I'm able to verify most of the changes, except those to `polyvec_compress()`, as there is no corresponding change in pq-crystals as far as I can tell. I'll just have to trust you that you got the magic constants right. (And one day I'll learn Kyber!)

boring-sys/patches/boring-pq.patch
bwesterb: From the name (taken from upstream) you would think `polyvec_compress` is just a component-wise version of `poly_vec`, which it's not. For Kyber768 the former leaves 10 bits per coefficient (the `d_u` parameter), whereas the latter leaves 4 (the `d_v` parameter).  Because the numbers are bigger, we also need a more accurate Barrett approximant for the division by Q. The intermediate values in the latter can grow as large as 2^42, whereas they stay below 32 in the former.

bwesterb:
Thanks for the review!

> I'm able to verify most of the changes, except those to `polyvec_compress()`, as there is no corresponding change in pq-crystals as far as I can tell. I'll just have to trust you that you got the magic constants right. (And one day I'll learn Kyber!)

The patch is [here](https://github.com/pq-crystals/kyber/commit/272125f6acc8e8b6850fd68ceb901a660ff48196)

boring-sys/patches/boring-pq.patch
```diff
@@ -20,43 +20,55 @@ This patch adds:
    key agreement should only be used for testing: to see if the smaller
    keyshare makes a difference.
 
+4. Supportfor IPDWing under codepoint 0xfe41. This key agreement
```
rushilmehra, line 23:
```suggestion
4. Support for IPDWing under codepoint 0xfe41. This key agreement
```

rushilmehra (approved):

inikulin (approved):

PR #248, opened 2024-07-23
Impl From for SslVersion

rushilmehra:


boring/src/ssl/mod.rs
```diff
@@ -598,6 +598,12 @@ impl SslVersion {
     pub const TLS1_3: SslVersion = SslVersion(ffi::TLS1_3_VERSION as _);
 }
 
+impl From<u16> for SslVersion {
```
inikulin, line 601: Should it be `TryFrom` that checks that version is valid?

inikulin (changes requested):

boring/src/ssl/mod.rs
rushilmehra: We do the same `impl From<u16>` for `ExtensionType` and `SslSignatureAlgorithm`, should I update those two as well? Although that would be a breaking change

rushilmehra: I swapped to `TryFrom`, and I can change the other instances of `From` if we're okay with the breaking change

inikulin: Ideally - yes. Let's probably not remove them just yet, but deprecate via an attribute

rushilmehra: It turns out `From` and `TryFrom` can't coexist: https://github.com/rust-lang/rust/issues/98062

Maybe we should just proceed as is and leave a warning in the very next release notes about this deprecation. Then in the next major bump we can fully swap to `TryFrom`

boring/src/ssl/mod.rs
```diff
@@ -625,6 +644,10 @@ impl fmt::Display for SslVersion {
 }
 
 /// A signature verification algorithm.
+///
+/// NOTE: The current implementation of `From` is unsound, as it's possible to create an
```
inikulin, line 648: Let's change it to `WARNING` in bold

inikulin (approved):

kornelski:
You could implement `try_from` as an inherent method (not a trait), and change `From` impl to use `ExtensionType::try_from(x).unwrap()` to make it sound.

rushilmehra:
My rationale is that if we went with the panic route, I consider that a breaking change, so at that point we should just switch to `TryFrom` directly. The issue with providing a default is it could lead to unexpected behavior. If we take `SslVersion` as an example, and you were able to construct one that is "invalid", the worst that would happen is any API like `set_min_proto_version` would just fail because the underlying c++ code sanity checks the incoming u16. If we provide a default, then application code that accidentally creates an `SslVersion` with a bogus value will silently configure some default version.

PR #252, opened 2024-08-01
Guard against empty strings given to select_next_proto

nox:


Lekensteyn (approved):

boring/src/ssl/test/mod.rs
```diff
@@ -275,6 +275,13 @@ fn test_alpn_server_select_none() {
     assert_eq!(None, s.ssl().selected_alpn_protocol());
 }
 
+#[test]
+fn empty_alpn() {
```
Lekensteyn, line 279: By convention, should this be:
```suggestion
fn test_empty_alpn() {
```

rushilmehra (approved):

Issue #260, opened 2024-08-05
git patch does not support UNC path


0x676e67:
Build test
```shell
cargo build --features rpk
```

Error Log

```log
The following warnings were emitted during compilation:

warning: boring-sys@4.9.1: applying RPK patch to boringssl

error: failed to run custom build command for `boring-sys v4.9.1 (D:\VSCode\boring\boring-sys)`

Caused by:
  process didn't exit successfully: `D:\VSCode\boring\target\debug\build\boring-sys-76d11f0ee260335a\build-script-main` (exit code: 101)
  --- stdout
  cargo:rerun-if-env-changed=BORING_BSSL_PATH_x86_64-pc-windows-msvc
  cargo:rerun-if-env-changed=BORING_BSSL_PATH_x86_64_pc_windows_msvc
  cargo:rerun-if-env-changed=HOST_BORING_BSSL_PATH
  cargo:rerun-if-env-changed=BORING_BSSL_PATH
  cargo:rerun-if-env-changed=BORING_BSSL_INCLUDE_PATH_x86_64-pc-windows-msvc
  cargo:rerun-if-env-changed=BORING_BSSL_INCLUDE_PATH_x86_64_pc_windows_msvc
  cargo:rerun-if-env-changed=HOST_BORING_BSSL_INCLUDE_PATH
  cargo:rerun-if-env-changed=BORING_BSSL_INCLUDE_PATH
  cargo:rerun-if-env-changed=BORING_BSSL_SOURCE_PATH_x86_64-pc-windows-msvc
  cargo:rerun-if-env-changed=BORING_BSSL_SOURCE_PATH_x86_64_pc_windows_msvc
  cargo:rerun-if-env-changed=HOST_BORING_BSSL_SOURCE_PATH
  cargo:rerun-if-env-changed=BORING_BSSL_SOURCE_PATH
  cargo:rerun-if-env-changed=BORING_BSSL_PRECOMPILED_BCM_O_x86_64-pc-windows-msvc
  cargo:rerun-if-env-changed=BORING_BSSL_PRECOMPILED_BCM_O_x86_64_pc_windows_msvc
  cargo:rerun-if-env-changed=HOST_BORING_BSSL_PRECOMPILED_BCM_O
  cargo:rerun-if-env-changed=BORING_BSSL_PRECOMPILED_BCM_O
  cargo:rerun-if-env-changed=BORING_BSSL_ASSUME_PATCHED_x86_64-pc-windows-msvc
[...]

PR #261, opened 2024-08-05
chore(boring-sys): Fix git apply patch on Windows

0x676e67:
close: https://github.com/cloudflare/boring/issues/260


rushilmehra:
This should be fine, but just to be safe can we gate that to `cfg[not(windows)]`? Also need to run `rustfmt`

rushilmehra (approved):

rushilmehra:
Merged, thanks!

PR #262, opened 2024-08-09
Expose mTLS related APIs, introduce ForeignTypeExt + ForeignTypeRefExt, add corresponds macro

rushilmehra:
`ForeignTypeExt` and `ForeignTypeRefExt` are inspired by https://github.com/sfackler/rust-openssl/pull/1345, which make dealing with FFI safer and more ergonomic. The new APIs also allow for gracefully handling instances where the initial API call results in `NULL`

### DIFF
--- a/boring/src/ssl/mod.rs
+++ b/boring/src/ssl/mod.rs
@@ -57,10 +57,10 @@
 //!     }
 //! }
 //! ```
-use crate::ffi;
 use foreign_types::{ForeignType, ForeignTypeRef, Opaque};
 use libc::{c_char, c_int, c_long, c_uchar, c_uint, c_void};
 use once_cell::sync::Lazy;
+use openssl_macros::corresponds;
 use std::any::TypeId;
 use std::collections::HashMap;
 use std::convert::TryInto;
@@ -82,6 +82,7 @@ use crate::dh::DhRef;
 use crate::ec::EcKeyRef;
 use crate::error::ErrorStack;
 use crate::ex_data::Index;
+use crate::ffi;
 use crate::nid::Nid;
 use crate::pkey::{HasPrivate, PKeyRef, Params, Private};
 use crate::srtp::{SrtpProtectionProfile, SrtpProtectionProfileRef};
@@ -243,9 +244,7 @@ pub struct SslMethod(*const ffi::SSL_METHOD);
 
 impl SslMethod {
     /// Support all versions of the TLS protocol.
-    ///
-    /// This corresponds to `TLS_method` on OpenSSL 1.1.0 and `SSLv23_method`
-    /// on OpenSSL 1.0.x.
+    #[corresponds(TLS_method)]
     pub fn tls() -> SslMethod {
         unsafe { SslMethod(TLS_method()) }
     }
@@ -257,25 +256,19 @@ impl SslMethod {
     }
 
     /// Support all versions of the DTLS protocol.
-    ///
-    /// This corresponds to `DTLS_method` on OpenSSL 1.1.0 and `DTLSv1_method`
-    /// on OpenSSL 1.0.x.
+    #[corresponds(DTLS_method)]
     pub fn dtls() -> SslMethod {
         unsafe { SslMethod(DTLS_method()) }
     }
 
     /// Support all versions of the TLS protocol, explicitly as a client.
-    ///
-    /// This corresponds to `TLS_client_method` on OpenSSL 1.1.0 and
-    /// `SSLv23_client_method` on OpenSSL 1.0.x.
+    #[corresponds(TLS_client_method)]
     pub fn tls_client() -> SslMethod {
         unsafe { SslMethod(TLS_client_method()) }
     }
 
     /// Support all versions of the TLS protocol, explicitly as a server.
-    ///
-    /// This corresponds to `TLS_server_method` on OpenSSL 1.1.0 and
-    /// `SSLv23_server_method` on OpenSSL 1.0.x.
+    #[corresponds(TLS_server_method)]
     pub fn tls_server() -> SslMethod {
         unsafe { SslMethod(TLS_server_method()) }
     }
@@ -285,6 +278,7 @@ impl SslMethod {
     /// # Safety
     ///
     /// The caller must ensure the pointer is valid.
+    #[corresponds(TLS_server_method)]
     pub unsafe fn from_ptr(ptr: *const ffi::SSL_METHOD) -> SslMethod {
         SslMethod(ptr)
     }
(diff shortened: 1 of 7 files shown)


boring/src/ssl/mod.rs
```diff
@@ -2975,6 +2994,27 @@ impl SslRef {
...
+        #[cfg(feature = "rpk")]
+        assert!(
+            !self.ssl_context().is_rpk(),
+            "This API is not supported for RPK"
+        );
+
+        unsafe {
+            let ptr = cert_store.as_ptr();
+            cvt(ffi::SSL_set0_verify_cert_store(self.as_ptr(), ptr) as c_int)?;
+            mem::forget(cert_store);
```
hcstern, line 2746: I took a brief look at the manpage and it's not clear to me why we need to pass cert_store by value and then forget it, rather than just passing a reference. Is it just that this only happens once on startup and it's okay to leak? It might be nice to have a [// SAFETY: block](https://std-dev-guide.rust-lang.org/policy/safety-comments.html)

boring/src/ssl/mod.rs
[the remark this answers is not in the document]
rushilmehra: Yeah given how difficult it is to get reviews, I highly prefer leaving everything in one PR and relying on the per-commit diff viewer. I can update the title to be more sensible

boring/src/ssl/mod.rs
hcstern: Also for some reason when I inspect the first commit individually it says the function as a whole is new, but when I look at master it's already there and looking at the second commit it only shows the actual cert_store.as_ptr() change. weird.

rushilmehra: On master you may be seeing the `SSL_CTX_set0_verify_cert_store` variant

rushilmehra: There are generally two variants for each API - one that operates on `SSL_CTX` (applies to the server generally), and another that operates on `SSL` (connection specific)

rushilmehra: Re: calling mem::forget manually, it's a good question, but I'm mainly following prior art: https://github.com/sfackler/rust-openssl/blob/master/openssl/src/ssl/mod.rs#L3408-L3414

rushilmehra: It's potentially something to do with the destructor being set to `X509_STORE_free`, in this case `mem::forget` will actually free memory without running the destructor. I haven't dug in fully, but I assume we don't want to run `X509_STORE_free` here

rushilmehra: The more I look into it the more I believe that to be the case, `SSL_set0_verify_cert_store` bumps the refcount on the underlying pointer and boringssl becomes the responsible party for managing that memory. If we let the user program `drop` the owned type, it will actually run `X509_STORE_free` which will cause issues. By calling `mem::forget`, we eliminate the possibility of the user messing with the `X509Store` object after instantiating it on boringssl state

hcstern: Agree on not wanting to free here due to the contract of verify_cert_store specified in the manpage, but it seems like the X509Store [has a pointer variant](https://docs.rs/boring/latest/boring/x509/store/struct.X509StoreRef.html) that we could use and then leave the decision of leaking vs keeping around in a context/state/whatever struct for reuse up to the caller.

But I do not have as much experience as sfackler so there's a good chance they had a reason for doing it this way.

rushilmehra: It does indeed have a pointer variant, but the `X509StoreBuilder` needs to yield an owned `X509Store` back to the caller before it's actually registered on boringssl state. If it yielded a reference, nothing would actually be keeping it alive under the hood. I think that's why all of the APIs that actually _register_ the store on the underlying boringssl state take ownership of the `X509Store` and call `mem::forget` on it

hcstern: To summarize: boringssl takes ownership of the `X509Store` (as Rushil mentioned above) and gets freed by boringssl. Thanks for taking the time to help me understand!

rushilmehra: For posterity, this works by virtue of refcounted objects in c++. When initialized, a `X509Store` will have a refcount of 1. Any time the destructor is called, the refcount will be decremented and checked against `== 0` before the object is fully freed. By using the `set0` variant of the API, we "hand off" the object at refcount == 1 to c++ and leave the memory's lifetime at mercy of the next destructor call

evanrittenhouse: Given that this is a common pattern I still think it would be nice to document somewhere

boring/src/x509/mod.rs
```diff
@@ -1044,6 +1044,36 @@ impl X509NameRef {
...
+    ///
+    /// This function will return `None` if the underlying string contains invalid utf-8.
+    #[corresponds(X509_NAME_oneline)]
+    pub fn oneline(&self) -> Option<String> {
+        unsafe {
+            let ptr = ffi::X509_NAME_oneline(self.as_ptr(), ptr::null_mut(), 0);
+            let out = CStr::from_ptr(ptr).to_str().map(|s| s.to_string());
+            // SAFETY: `X509_NAME_oneline` dynamically allocates memory for the string that's
+            // ultimately returned. As such, it's the caller's responsibility to free that memory
+            // after allocating an owned String in rust land.
```
evanrittenhouse, line 1057: Should this be in the doc comment?

boring/src/ssl/mod.rs
```diff
@@ -3800,11 +3441,24 @@ impl SslRef {
...
+    #[corresponds(SSL_set_client_CA_list)]
+    pub fn set_client_ca_list(&mut self, list: Stack<X509Name>) {
+        #[cfg(feature = "rpk")]
+        assert!(
+            !self.ssl_context().is_rpk(),
+            "This API is not supported for RPK"
+        );
+
+        unsafe { ffi::SSL_set_client_CA_list(self.as_ptr(), list.as_ptr()) }
+        mem::forget(list);
```
bwesterb, line 3457: Why is this legitimate?

rushilmehra: `SSL_set_client_CA_list` takes ownership of the list: https://github.com/google/boringssl/blob/master/include/openssl/ssl.h#L2966

Constructing the `Stack<X509Name>` requires the application to own the `Stack`, and the owned type will free the underlying memory if `drop` is called. The general pattern here is to pass ownership to boringssl, and call `mem::forget` to release that ownership without calling `drop` and destroying the underlying stack

bwesterb (approved):

boring/src/ssl/mod.rs
nox: The intent should be conveyed by using [`<X509Store as ForeignType>::into_ptr`](https://docs.rs/foreign-types-shared/0.3.1/foreign_types_shared/trait.ForeignType.html#method.into_ptr)

rushilmehra: Opened https://github.com/cloudflare/boring/pull/267

PR #264, opened 2024-08-16
Prepare for syncing with upstream

rushilmehra:
* Support linking with a runtime cpp library
> As of https://boringssl-review.googlesource.com/c/boringssl/+/66288,
libssl allows a C++ runtime dependency. As such, we need to link with a
cpp runtime library. Implementation is inspired heavily from
https://github.com/google/boringssl/commit/54c956b2e668e11c75f1ee0367f1b3a0ad28eff9. Before releasing this change, we'll need to figure out a way to support
this for windows.

* Fix bug with accessing memzero'd X509StoreContext in tests
> As of https://boringssl-review.googlesource.com/c/boringssl/+/64141,
X509_STORE_CTX_cleanup will zero the memory allocated to the
X509_STORE_CTX. Because X509StoreContextRef::init invokes
X509_STORE_CTX_cleanup once the with_context closure has finished,
calling X509StoreContextRef::verify_result (or any API really) is going
to be invalid because memory has been zerod out. This is a pretty big
footgun, so maybe we should consider screaming a bit louder for this
case.

This patch doesn't sync with upstream _yet_, because all of our patches break due to upstream directory structure changes, and the RPK patch needs to be completely reworked

### DIFF
--- a/boring-sys/build/config.rs
+++ b/boring-sys/build/config.rs
@@ -34,6 +34,7 @@ pub(crate) struct Env {
     pub(crate) opt_level: Option<OsString>,
     pub(crate) android_ndk_home: Option<PathBuf>,
     pub(crate) cmake_toolchain_file: Option<PathBuf>,
+    pub(crate) cpp_runtime_lib: Option<OsString>,
 }
 
 impl Config {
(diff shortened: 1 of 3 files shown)


bwesterb (approved):

kornelski (approved):

RvdE:
This patch is causing build failure on FreeBSD. Not sure about the other BSD's, but FreeBSD doesn't have libstdc++, but does have libc++.

PR #266, opened 2024-09-02
Expose `SSL_CTX_set_info_callback`

evanrittenhouse:


boring/src/ssl/callbacks.rs
```diff
@@ -521,3 +521,25 @@ where
...
+) where
+    F: Fn(&SslRef, i32, i32) + Send + Sync + 'static,
+{
+    // Due to FFI signature requirements we have to pass a *const SSL into this function, but
+    // foreign-types requires a *mut SSL to get the Rust SslRef
+    let mut_ref = ssl as *mut ffi::SSL;
+
+    // SAFETY: boring provides valid inputs.
+    let ssl = unsafe { SslRef::from_ptr(mut_ref) };
+    let ssl_context = ssl.ssl_context().to_owned();
```
nox, line 538: Don't need `to_owned` here as `callback` take a `&SslRef` not a `&mut SslRef`.

boring/src/ssl/mod.rs
```diff
@@ -834,6 +834,47 @@ pub fn select_next_proto<'a>(server: &[u8], client: &'a [u8]) -> Option<&'a [u8]
     }
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct Ssl3AlertLevel(i32);
+
+impl Ssl3AlertLevel {
```
nox, line 840: This does not seem to follow the idiom used by the various ints provided by boring. The others have `Self` constants that wrap the `ffi` values.

boring/src/ssl/callbacks.rs
```diff
@@ -521,3 +521,25 @@ where
         Err(err) => err.0,
     }
 }
+
+pub(super) unsafe extern "C" fn raw_info_callback<F>(
+    ssl: *const ffi::SSL,
+    mode: c_int,
+    value: c_int,
+) where
+    F: Fn(&SslRef, i32, i32) + Send + Sync + 'static,
```
nox, line 530: Shouldn't those `i32` be `Ssl3AlertLevel` and `SslInfoCallbackMode`

evanrittenhouse: The latter `i32` can be a variety of different things depending on the `mode`, I'll encode that in an enum with some helpers to make it more ergonomic for the user

boring/src/ssl/mod.rs
```diff
@@ -834,6 +834,77 @@ pub fn select_next_proto<'a>(server: &[u8], client: &'a [u8]) -> Option<&'a [u8]
...
+    pub const WARNING: Ssl3AlertLevel = Self(ffi::SSL3_AL_WARNING);
+    pub const FATAL: Ssl3AlertLevel = Self(ffi::SSL3_AL_FATAL);
+}
+
+impl TryFrom<i32> for Ssl3AlertLevel {
+    type Error = &'static str;
+
+    fn try_from(value: i32) -> Result<Self, Self::Error> {
+        match value >> 8 {
```
evanrittenhouse, line 900: Do we want to add an `assert` here instead of the catch-all branch given that this comes from Boring?

nox: I would just remove the impl and make the bindings do `Ssl3AlertLevel(value)` when passing the value to user code, if boring returns another value, so be it, it may just be a value that we did not write a binding for yet.

nox (approved):

boring/src/ssl/mod.rs
evanrittenhouse: Since the level is the first byte of the u16, I've changed it to pass `SslAlertLevel(self.0 >> 8)`, similar to how the actual alert value bitmasks on 256

Issue #285, opened 2024-10-17
MacOS FIPS crossbuild CI runner is broken


evanrittenhouse:
Looks like:

```
warning: boring-sys@4.10.3: Compiler family detection failed due to error: ToolNotFound: Failed to find tool. Is `x86_64-unknown-linux-gnu-gcc` installed?
[221](https://github.com/cloudflare/boring/actions/runs/11025392135/job/30620183456#step:11:222)warning: boring-sys@4.10.3: Compiler family detection failed due to error: ToolNotFound: Failed to find tool. Is `x86_64-unknown-linux-gnu-g++` installed?
[222](https://github.com/cloudflare/boring/actions/runs/11025392135/job/30620183456#step:11:223)warning: boring-sys@4.10.3: Compiler family detection failed due to error: ToolNotFound: Failed to find tool. Is `x86_64-unknown-linux-gnu-gcc` installed?
```

First started failing in https://github.com/cloudflare/boring/actions/runs/11025392135/job/30620183456

PR #286, opened 2024-10-17
Add "fips-compat" feature

ghedo:
This adds a feature to build against a BoringSSL version compatible with the current boringssl-fips, but _without_ actually enabling the `fips` feature.

This can be useful to use with `fips-link-precompiled` while using a custom BoringSSL version based on the older FIPS branch.


rushilmehra (approved):

rushilmehra:
FYI failing build is due to https://github.com/cloudflare/boring/issues/285

PR #287, opened 2024-10-21
Add `set_cert_verify_callback` (`SSL_CTX_set_cert_verify`)

semaj-cf:
Add a wrapper for `SSL_CTX_set_cert_verify`, which allows consumers to override the default certificate verification behavior.

The binding resembles `SSL_CTX_set_verify`'s.

See
https://docs.openssl.org/master/man3/SSL_CTX_set_cert_verify_callback/ for more details.

### DIFF
--- a/boring/src/ssl/callbacks.rs
+++ b/boring/src/ssl/callbacks.rs
@@ -64,6 +64,34 @@ where
     unsafe { raw_custom_verify_callback(ssl, out_alert, callback) }
 }
 
+pub(super) unsafe extern "C" fn raw_cert_verify<F>(
+    x509_ctx: *mut ffi::X509_STORE_CTX,
+    _arg: *mut c_void,
+) -> c_int
+where
+    F: Fn(&mut X509StoreContextRef) -> bool + 'static + Sync + Send,
+{
+    // SAFETY: boring provides valid inputs.
+    let ctx = unsafe { X509StoreContextRef::from_ptr_mut(x509_ctx) };
+
+    let ssl_idx = X509StoreContext::ssl_idx().expect("BUG: store context ssl index missing");
+    let verify_idx = SslContext::cached_ex_index::<F>();
+
+    let verify = ctx
+        .ex_data(ssl_idx)
+        .expect("BUG: store context missing ssl")
+        .ssl_context()
+        .ex_data(verify_idx)
+        .expect("BUG: verify callback missing");
+
+    // SAFETY: The callback won't outlive the context it's associated with
+    // because there is no way to get a mutable reference to the `SslContext`,
+    // so the callback can't replace itself.
+    let verify = unsafe { &*(verify as *const F) };
+
+    verify(ctx) as c_int
+}
+
 pub(super) unsafe extern "C" fn ssl_raw_custom_verify<F>(
     ssl: *mut ffi::SSL,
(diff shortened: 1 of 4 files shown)


semaj-cf:
failing build looks related to: https://github.com/cloudflare/boring/issues/285

boring/src/ssl/mod.rs
```diff
@@ -1000,6 +1000,40 @@ impl SslContextBuilder {
...
+    ///
+    /// *Warning*: Providing a complete verification procedure is a complex task. See
+    /// https://docs.openssl.org/master/man3/SSL_CTX_set_cert_verify_callback/#notes for more
+    /// information.
+    ///
+    /// # Panics
+    ///
+    /// This method panics if this `SslContext` is associated with a RPK context.
+    #[corresponds(SSL_CTX_set_cert_verify_callback)]
+    pub fn set_cert_verify_callback<F>(&mut self, callback: F)
```
bwesterb, line 1020: You can unset by passing NULL, but that's not possible with this binding.

boring/src/ssl/mod.rs
```diff
@@ -1000,6 +1000,40 @@ impl SslContextBuilder {
...
+    where
+        F: Fn(&mut X509StoreContextRef) -> bool + 'static + Sync + Send,
+    {
+        #[cfg(feature = "rpk")]
+        assert!(!self.is_rpk, "This API is not supported for RPK");
+
+        self.replace_ex_data(SslContext::cached_ex_index::<F>(), callback);
```
bwesterb, line 1027: `set_verify_callback` uses

```
            // this needs to be in an Arc since the callback can register a new callback!
            self.replace_ex_data(Ssl::cached_ex_index(), Arc::new(callback)); 
```

semaj-cf: from my digging, this is copy-pasta. see this comment by @nox https://github.com/cloudflare/boring/blob/master/boring/src/ssl/callbacks.rs#L200-L202

I agree that it is impossible to get a mutable reference to an `SslRef` or `SslContextRef`, so I don't see why this needs to be an `Arc`. I am happy to make it one just to be safe, though

boring/src/ssl/mod.rs
semaj-cf: it seems we either change this to `Option` or add another function that unsets this callback. Changing this to `Option` seems to make Rust unable to reason about arguments to this function when it is called, requiring type annotations (which are onerous when you just want to pass `None`).

semaj-cf: added a `TODO`

boring/src/ssl/mod.rs
bwesterb: Add a comment with the reference.

bwesterb (approved):

boring/src/ssl/callbacks.rs
```diff
@@ -64,6 +64,34 @@ where
     unsafe { raw_custom_verify_callback(ssl, out_alert, callback) }
 }
 
+pub(super) unsafe extern "C" fn cert_verify<F>(
```
rushilmehra, line 67:
```suggestion
pub(super) unsafe extern "C" fn raw_cert_verify<F>(
```

PR #297, opened 2024-11-26
x509 verify params + safety fixes

kornelski:
I've copied rust-openssl APIs that are needed to make c2pa.org SDK compatible with boring (jira IMAGES-1371).

The openssl crate had also a couple of important safety fixes — `X509VerifyParam::set_host` with 0-len host was unsafe, and `X509StoreRef::objects` references may get invalidated by other accesses of the shared cache. 

`X509StoreRef::objects` is unfixable, so I've marked it as `unsafe` and deprecated, which is technically semver-breaking, but I think the risk of use-after-free in a safe API could warrant releasing it in 4.x despite breakage.

### DIFF
--- a/boring/src/x509/verify.rs
+++ b/boring/src/x509/verify.rs
@@ -1,15 +1,16 @@
 use crate::ffi;
-use foreign_types::ForeignTypeRef;
-use libc::{c_uint, c_ulong};
+use foreign_types::{ForeignType, ForeignTypeRef};
+use libc::{c_int, c_uint, c_ulong, time_t};
 use std::net::IpAddr;
 use openssl_macros::corresponds;
 
-use crate::cvt;
 use crate::error::ErrorStack;
+use crate::{cvt, cvt_p};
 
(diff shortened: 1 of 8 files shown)


boring/src/x509/verify.rs
```diff
@@ -77,19 +102,43 @@ impl X509VerifyParamRef {
         }
     }
 
+    /// Gets verification flags.
+    #[corresponds(X509_VERIFY_PARAM_get_flags)]
+    pub fn flags(&mut self) -> X509VerifyFlags {
```
nox, line 115: Why is this `&mut self`?

kornelski: That's what openssl-rust has, and bindgen generates `pub fn X509_VERIFY_PARAM_get_flags(param: *mut X509_VERIFY_PARAM) -> ::std::os::raw::c_ulong;` with a `mut` pointer.

rushilmehra (approved):

PR #304, opened 2024-12-06
Refactor!: Introduce a Cargo feature for optional Hyper 0 support

### LINKED ISSUE #294: Make Hyper v0 support optional
Dear maintainers,

Now that Hyper v1 has been supported for a few versions and that it works fine, it seems like a good time to consider retiring v0. Currently, only adding v1 is possible without forking, thus doubling the hyper stack's dependencies instead of replacing them, since v0 is always enabled even though unused.

Would it therefore be possible to either remove v0 or at least make it optional? Making the `v0` module get behind `cfg(not(feature = "hyper1"))` seems to work. The idea would be to have the full re-export kept for both modules so that only the Cargo feature controls what is selected and passing from one to the other would be a drop-in, at least from the item paths point of view. Not pulling any v0 when v1 is enabled and vice-versa would be a priority.

I've made an attempt at this in #304.

Thanks in advance,
Paul.

PaulDance:
Closes #294. Requires breaking changes, but only regarding the `hyper1` feature's enabled module path; the default v0 is still preserved. What is proposed here in order to achieve the objective is to have both modules be re-exported when the other feature is not enabled, and have them be made public when both features are enabled at the same time. This way, the features should be kept additive: all four combinations are supported. It enables dependency de-duplication when consuming it.

### DIFF
--- a/hyper-boring/src/lib.rs
+++ b/hyper-boring/src/lib.rs
@@ -11,12 +11,14 @@ use std::fmt;
 use tokio_boring::SslStream;
 
 mod cache;
-mod v0;
-/// Hyper 1 support.
+/// Hyper 0 support.
+#[cfg(feature = "hyper0")]
+pub mod v0;
 #[cfg(feature = "hyper1")]
-pub mod v1;
+mod v1;
 
-pub use self::v0::*;
+#[cfg(feature = "hyper1")]
+pub use self::v1::*;
 
 fn key_index() -> Result<Index<Ssl, SessionKey>, ErrorStack> {
     static IDX: OnceCell<Index<Ssl, SessionKey>> = OnceCell::new();
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -366,5 +366,7 @@ jobs:
       name: Run `rpk,underscore-wildcards` tests
     - run: cargo test --features pq-experimental,rpk,underscore-wildcards
       name: Run `pq-experimental,rpk,underscore-wildcards` tests
-    - run: cargo test -p hyper-boring --features hyper1
+    - run: cargo test -p hyper-boring --features hyper1-runtime
       name: Run hyper 1.0 tests for hyper-boring
+    - run: cargo test -p hyper-boring --features hyper0-runtime
+      name: Run hyper 0. tests for hyper-boring
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,12 +37,12 @@ futures = "0.3"
 tokio = "1"
 anyhow = "1"
 antidote = "1.0.0"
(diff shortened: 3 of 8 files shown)


hyper-boring/src/lib.rs
```diff
@@ -11,12 +11,21 @@ use std::fmt;
...
+mod v1;
 /// Hyper 1 support.
-#[cfg(feature = "hyper1")]
+#[cfg(all(feature = "hyper1", feature = "hyper0"))]
 pub mod v1;
 
+#[cfg(all(feature = "hyper0", not(feature = "hyper1")))]
 pub use self::v0::*;
+#[cfg(all(feature = "hyper1", not(feature = "hyper0")))]
+pub use self::v1::*;
```
kornelski, line 28: This means the features aren't purely additive, and could create confusing type errors if both flags end up being set by accident. Maybe always keep `v0` in a module?

PaulDance: > This means the features aren't purely additive, and could create confusing type errors if both flags end up being set by accident.

To be clear, the objective of the `#[cfg]` gate is to properly take that case into consideration: if both are set, then the items are kept in explicit module paths, but if only one is set, all items are re-exported in order to keep a simpler pathing in what should be most cases I imagine. This should help avoid type collisions like you mentioned. Couldn't this work you think?

> Maybe always keep `v0` in a module?

I can do that, but it would mean breaking API even more: current paths are `crate::HttpsConnector` for the v0 and `crate::v1::HttpsConnector` for the v1, while inverting like you suggest would do `crate::v0::HttpsConnector` for the v0 and crate::HttpsConnector` for the v1, I guess. Is this still acceptable?

kornelski: The names of features have changed, so this will need to be a semver-major release anyway.

I am concerned about the items re-exported at the root, because they change in non-additive way. The feature flags are okay if users only used `v0::…`/`v1::…` types, but the re-export can be a problem if users mix use of the top-level items with `v0::…` items, or top-level items with another crate.

Re-exporting only v1 types at the top level would be more robust.

PaulDance: Alright by me! I was just trying not to break too much if possible.

Also, on top of that and since it seems we go more in the direction of v1, I've made it so v1 is now the default and v0 is not anymore but still available through the same features.

kornelski (approved):

PaulDance:
Done! Also made it so that exclusive v0 users avoid pulling Tower uselessly by moving features around a bit, which I forgot to do in the previous PR.

kornelski (approved):

PaulDance:
Works like a charm downstream, many thanks!

PR #314, opened 2025-02-19
Expose SSL_CTX_set1_ech_keys from SslContextRef

rushilmehra:
We currently expose this method on `SslContextBuilder`, which is fine for bootstrapping an `SSL_CTX`, but subsequent attempts to set ECH keys (like during key rotation) can only happen via `SslContextRef`. Also update the method on the builder to take an immutable reference to self because the API is thread safe.

### DIFF
--- a/boring/src/ssl/mod.rs
+++ b/boring/src/ssl/mod.rs
@@ -1960,7 +1960,7 @@ impl SslContextBuilder {
     /// threads.
     #[cfg(not(feature = "fips"))]
     #[corresponds(SSL_CTX_set1_ech_keys)]
-    pub fn set_ech_keys(&mut self, keys: SslEchKeys) -> Result<(), ErrorStack> {
+    pub fn set_ech_keys(&self, keys: SslEchKeys) -> Result<(), ErrorStack> {
         unsafe { cvt(ffi::SSL_CTX_set1_ech_keys(self.as_ptr(), keys.as_ptr())).map(|_| ()) }
     }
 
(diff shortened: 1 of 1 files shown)


boring/src/ssl/mod.rs
```diff
@@ -1960,7 +1960,7 @@ impl SslContextBuilder {
     /// threads.
     #[cfg(not(feature = "fips"))]
     #[corresponds(SSL_CTX_set1_ech_keys)]
-    pub fn set_ech_keys(&mut self, keys: SslEchKeys) -> Result<(), ErrorStack> {
+    pub fn set_ech_keys(&self, keys: SslEchKeys) -> Result<(), ErrorStack> {
         unsafe { cvt(ffi::SSL_CTX_set1_ech_keys(self.as_ptr(), keys.as_ptr())).map(|_| ()) }
```
bwesterb, line 1964: Why doesn't this take a mut ref?

bwesterb: I suppose it's fine as we're just passing an opaque value into OpenSSL which handles the allocation, right?

rushilmehra: This API is meant to be called after the SSL_CTX has presumably been distributed to multiple threads, so it's thread safe under the hood. 

We take ownership of `keys` and drop it at the end of the function to shed our refcount and let boringssl take care of the rest

bwesterb (approved):

PR #319, opened 2025-02-21
Revert breaking changes

rushilmehra:
Clippy lints gettiing fixed in https://github.com/cloudflare/boring/pull/318


kornelski (approved):

PaulDance:
Why this? There is no real justification for such a big revert of a previously-accepted change. Would it be possible to introduce it back now, maybe in a backwards-compatible way instead?

PR #321, opened 2025-02-21
Revert cmake bump (for now) as it is overly restrictive

rushilmehra:
Some users of boring have issues with newer versions of cmake. Because we have an alternative solution, we can hold off on the bump for now.


rushilmehra:
cc @jrose-signal

kornelski (approved):

PR #329, opened 2025-03-10
Advertise X25519MLKEM768 with "kx-client-pq-preferred"

cjpatton:
This algorithm is advertised with "kx-client-pq-supported" but not with "preferred". However the algorithm is wide spread enough that preferring it is not a significant risk.


bwesterb (approved):

rushilmehra:
Rebase should fix the CI failure

cjpatton:
Rebased!

PR #330, opened 2025-03-11
Add set_ticket_key_callback (SSL_CTX_set_tlsext_ticket_key_cb)

toidiu:
Add a wrapper for the `SSL_CTX_set_tlsext_ticket_key_cb`, which allows consumers to configure the EVP_CIPHER_CTX and HMAC_CTX used for encrypting/decrypting session tickets.

See https://docs.openssl.org/1.0.2/man3/SSL_CTX_set_tlsext_ticket_key_cb/ for more details.

### DIFF
--- a/boring/src/hmac.rs
+++ b/boring/src/hmac.rs
@@ -14,11 +14,7 @@ impl HmacCtxRef {
     /// Configures HmacCtx to use `md` as the hash function and `key` as the key.
     ///
     /// https://commondatastorage.googleapis.com/chromium-boringssl-docs/hmac.h.html#HMAC_Init_ex
-    ///
-    /// # Safety
-    ///
-    /// The caller must ensure HMAC_CTX has been initalized.
-    pub unsafe fn init(&mut self, key: &[u8], md: &MessageDigest) -> Result<(), ErrorStack> {
+    pub fn init(&mut self, key: &[u8], md: &MessageDigest) -> Result<(), ErrorStack> {
         ffi::init();
 
         unsafe {
--- a/boring/src/hmac.rs
+++ b/boring/src/hmac.rs
@@ -1,9 +1,7 @@
 use crate::cvt;
 use crate::error::ErrorStack;
+use crate::foreign_types::ForeignTypeRef;
 use crate::hash::MessageDigest;
-use std::ffi::c_void;
-
-use foreign_types::ForeignType;
 
 foreign_type_and_impl_send_sync! {
     type CType = ffi::HMAC_CTX;
@@ -12,7 +10,7 @@ foreign_type_and_impl_send_sync! {
     pub struct HmacCtx;
 }
 
-impl HmacCtx {
+impl HmacCtxRef {
     /// Configures HmacCtx to use `md` as the hash function and `key` as the key.
     ///
     /// https://commondatastorage.googleapis.com/chromium-boringssl-docs/hmac.h.html#HMAC_Init_ex
@@ -26,7 +24,7 @@ impl HmacCtx {
         unsafe {
             cvt(ffi::HMAC_Init_ex(
                 self.as_ptr(),
-                key.as_ptr() as *const c_void,
+                key.as_ptr().cast(),
                 key.len(),
                 md.as_ptr(),
                 // ENGINE api is deprecated
--- a/boring/src/hmac.rs
+++ b/boring/src/hmac.rs
@@ -1,0 +1,38 @@
+use crate::cvt;
+use crate::error::ErrorStack;
+use crate::hash::MessageDigest;
+use std::ffi::c_void;
+
+use foreign_types::ForeignType;
+
+foreign_type_and_impl_send_sync! {
+    type CType = ffi::HMAC_CTX;
+    fn drop = ffi::HMAC_CTX_free;
+
+    pub struct HmacCtx;
+}
+
+impl HmacCtx {
+    /// Configures HmacCtx to use `md` as the hash function and `key` as the key.
+    ///
+    /// https://commondatastorage.googleapis.com/chromium-boringssl-docs/hmac.h.html#HMAC_Init_ex
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure HMAC_CTX has been initalized.
+    pub unsafe fn init(&mut self, key: &[u8], md: &MessageDigest) -> Result<(), ErrorStack> {
+        ffi::init();
+
+        unsafe {
+            cvt(ffi::HMAC_Init_ex(
+                self.as_ptr(),
+                key.as_ptr() as *const c_void,
+                key.len(),
+                md.as_ptr(),
+                // ENGINE api is deprecated
+                core::ptr::null_mut(),
+            ))
+            .map(|_| ())
+        }
+    }
+}
--- a/boring/src/ssl/callbacks.rs
+++ b/boring/src/ssl/callbacks.rs
@@ -299,7 +299,7 @@ where
         .ex_data::<F>(SslContext::cached_ex_index::<F>())
         .expect("expected session resumption callback");
 
-    // Safety: the callback guarantees that key_name is 16 bytes
+    // SAFETY: the callback guarantees that key_name is 16 bytes
     let key_name =
         unsafe { slice::from_raw_parts_mut(key_name, ffi::SSL_TICKET_KEY_NAME_LEN as usize) };
     let key_name = <&mut [u8; 16]>::try_from(key_name).expect("boring provides a 16-byte key name");
--- a/boring/src/ssl/callbacks.rs
+++ b/boring/src/ssl/callbacks.rs
@@ -282,7 +282,7 @@ where
     F: Fn(
             &SslRef,
             &mut [u8; 16],
-            *mut u8,
+            &mut [u8; ffi::EVP_MAX_IV_LENGTH as usize],
             *mut ffi::EVP_CIPHER_CTX,
             *mut ffi::HMAC_CTX,
             bool,
@@ -304,8 +304,22 @@ where
         unsafe { slice::from_raw_parts_mut(key_name, ffi::SSL_TICKET_KEY_NAME_LEN as usize) };
     let key_name = <&mut [u8; 16]>::try_from(key_name).expect("boring provides a 16-byte key name");
 
+    // SAFETY: the callback provides 16 bytes iv
+    //
+    // https://github.com/google/boringssl/blob/main/ssl/ssl_session.cc#L331
+    let iv = unsafe { core::slice::from_raw_parts_mut(iv, ffi::EVP_MAX_IV_LENGTH as usize) };
+    let iv = <&mut [u8; ffi::EVP_MAX_IV_LENGTH as usize]>::try_from(iv)
+        .expect("boring provides a 16-byte iv");
+
     // When encrypting a new ticket, encrypt will be one.
     let encrypt = encrypt == 1;
+
+    // Zero-initialize the key_name and iv, since the application is expected to populate these
+    // fields in the encrypt mode.
+    if encrypt {
+        unsafe { ptr::write(key_name, [0; 16]) };
+        unsafe { ptr::write(iv, [0; ffi::EVP_MAX_IV_LENGTH as usize]) };
+    }
 
     callback(ssl, key_name, iv, evp_ctx, hmac_ctx, encrypt).into()
 }
--- a/boring/src/ssl/callbacks.rs
+++ b/boring/src/ssl/callbacks.rs
@@ -12,9 +12,9 @@ use crate::ssl::TicketKeyCallbackResult;
 use crate::x509::{X509StoreContext, X509StoreContextRef};
 use foreign_types::ForeignType;
 use foreign_types::ForeignTypeRef;
-use libc::c_char;
-use libc::{c_int, c_uchar, c_uint, c_void};
+use libc::{c_char, c_int, c_uchar, c_uint, c_void};
 use std::ffi::CStr;
+use std::mem::MaybeUninit;
 use std::ptr;
 use std::slice;
 use std::str;
@@ -270,6 +270,11 @@ where
     }
 }
 
+unsafe fn to_uninit<'a, T: 'a>(ptr: *mut T) -> &'a mut MaybeUninit<T> {
+    assert!(!ptr.is_null());
+    unsafe { &mut *ptr.cast::<MaybeUninit<T>>() }
+}
+
 pub(super) unsafe extern "C" fn raw_ticket_key<F>(
     ssl: *mut ffi::SSL,
     key_name: *mut u8,
@@ -301,15 +306,12 @@ where
 
     // SAFETY: the callback guarantees that key_name is 16 bytes
     let key_name =
-        unsafe { slice::from_raw_parts_mut(key_name, ffi::SSL_TICKET_KEY_NAME_LEN as usize) };
-    let key_name = <&mut [u8; 16]>::try_from(key_name).expect("boring provides a 16-byte key name");
+        unsafe { to_uninit(key_name.cast::<[u8; ffi::SSL_TICKET_KEY_NAME_LEN as usize]>()) };
 
     // SAFETY: the callback provides 16 bytes iv
     //
     // https://github.com/google/boringssl/blob/main/ssl/ssl_session.cc#L331
-    let iv = unsafe { core::slice::from_raw_parts_mut(iv, ffi::EVP_MAX_IV_LENGTH as usize) };
-    let iv = <&mut [u8; ffi::EVP_MAX_IV_LENGTH as usize]>::try_from(iv)
-        .expect("boring provides a 16-byte iv");
+    let iv = unsafe { to_uninit(iv.cast::<[u8; ffi::EVP_MAX_IV_LENGTH as usize]>()) };
 
     // When encrypting a new ticket, encrypt will be one.
     let encrypt = encrypt == 1;
@@ -317,9 +319,11 @@ where
     // Zero-initialize the key_name and iv, since the application is expected to populate these
     // fields in the encrypt mode.
     if encrypt {
-        unsafe { ptr::write(key_name, [0; 16]) };
-        unsafe { ptr::write(iv, [0; ffi::EVP_MAX_IV_LENGTH as usize]) };
+        *key_name = MaybeUninit::zeroed();
+        *iv = MaybeUninit::zeroed();
     }
+    let key_name = unsafe { key_name.assume_init_mut() };
+    let iv = unsafe { iv.assume_init_mut() };
 
     callback(ssl, key_name, iv, evp_ctx, hmac_ctx, encrypt).into()
 }
--- a/boring/src/ssl/callbacks.rs
+++ b/boring/src/ssl/callbacks.rs
@@ -8,13 +8,15 @@ use super::{
 };
 use crate::error::ErrorStack;
 use crate::ffi;
+use crate::hmac::HmacCtx;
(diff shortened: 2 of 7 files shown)


boring/src/ssl/callbacks.rs
```diff
@@ -269,6 +270,46 @@ where
...
+{
+    // SAFETY: boring provides valid inputs.
+    let ssl = unsafe { SslRef::from_ptr_mut(ssl) };
+
+    let ssl_context = ssl.ssl_context().to_owned();
+    let callback = ssl_context
+        .ex_data::<F>(SslContext::cached_ex_index::<F>())
+        .expect("expected session resumption callback");
+
+    // Safety: the callback guarantees that key_name is 16 bytes
```
rushilmehra, line 302:
```suggestion
    // SAFETY: the callback guarantees that key_name is 16 bytes
```

boring/src/ssl/mod.rs
```diff
@@ -882,6 +882,50 @@ pub enum SslInfoCallbackValue {
...
+pub enum TicketKeyCallbackResult {
+    /// Abort the handshake.
+    Error,
+
+    /// The peer supplied session ticket was not recognized. Continue with a full handshake.
+    ///
+    /// # Note
+    ///
+    /// This is a decryption specific status code.
+    DecryptTicketUnrecognized,
```
rushilmehra, line 896: Technically decryption can fail due to a server-side issue, so might be better to name this variant `DecryptTicketFailed`

boring/src/ssl/test/session_resumption.rs
```diff
@@ -0,0 +1,159 @@
...
+    let mut server = Server::builder();
+    server.expected_connections_count(2);
+    let server = server.build();
+
+    let mut client = server.client();
+    client
+        .ctx()
+        .set_session_cache_mode(SslSessionCacheMode::CLIENT);
+    client.ctx().set_new_session_callback(|_, session| {
+        let _can_receive_multiple_tickets = SESSION_TICKET.set(session.to_der().unwrap());
```
rushilmehra, line 28:
```suggestion
        SESSION_TICKET.set(session.to_der().unwrap());
```

boring/src/ssl/mod.rs
```diff
@@ -1164,6 +1208,43 @@ impl SslContextBuilder {
...
+                *mut ffi::EVP_CIPHER_CTX,
+                *mut ffi::HMAC_CTX,
+                bool,
+            ) -> TicketKeyCallbackResult
+            + 'static
+            + Sync
+            + Send,
+    {
+        #[cfg(feature = "rpk")]
+        assert!(!self.is_rpk, "This API is not supported for RPK");
```
toidiu, line 1240: I copied this from the other callbacks and not sure if it applies here. What is RPK and does it apply for this callback?

boring/src/ssl/test/session_resumption.rs
toidiu: `OnceLock:set()` returns a Result which we need to consume. I'll refactor to add a comment instead.

> Returns Ok(()) if the cell's value was set by this call.

boring/src/ssl/mod.rs
rushilmehra: RPK is raw public keys (https://datatracker.ietf.org/doc/html/rfc7250), it's fine to leave these callbacks unimplemented there

boring/src/ssl/mod.rs
toidiu: > In this mode, callback returns -1 to abort the handshake, 0 if the ticket key was unrecognized, and 1 or 2 on success. If it returns 2, the ticket will be renewed.

I got unrecognized wording from the [boring docs](https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_CTX_set_tlsext_ticket_key_cb) and took it to mean that the server couldn't find the ticket when it tried to lookup the `key_name` in its store. Shouldn't a server-side issue be an Error?

> In this mode, callback returns 1 on success, 0 to decline sending a ticket, and -1 on error.

Also since we are on an older version of boring, the "0 to decline sending a ticket" code doesn't apply to our version yet. Once we upgrade, this name should be change to be less Decrypting specific. I am not familiar how liberally we handle breaking changes in this repo and wonder if we should rename this to be more future proof?

boring/src/ssl/test/session_resumption.rs
rushilmehra: You can just unwrap it since this is tests

boring/src/ssl/mod.rs
rushilmehra: Yeah interestingly boringssl changed the documentation ~4 months ago: 
* https://boringssl-review.googlesource.com/c/boringssl/+/73027
* https://boringssl-review.googlesource.com/c/boringssl/+/73028

Boringssl breaks their API at will, but we follow semver for our public API, so it's best to future-proof as much as possible

boring/src/ssl/test/session_resumption.rs
toidiu: Since we receive 2 NST from the server, [OnceLock::set](https://doc.rust-lang.org/beta/std/sync/struct.OnceLock.html#method.set).unwrap() will succeed for the 1st ticket but fail for the 2nd one. Since I just care to retrieve any NST, I need to ignore the result here.

Alternatively, I could do a `OnceLock::get().is_some()` check before `set`ting the value, which would avoid an Error value.

toidiu: I update the logic and added checks for the expected behavior.

boring/src/ssl/mod.rs
toidiu: > Boringssl breaks its API at will, but we follow semver for our public API, so it's best to future-proof as much as possible

I updated the the status code to `Noop` which seems appropriate for both cases in the future.

Updated in https://github.com/cloudflare/boring/pull/330/commits/2a43889aaa1607d112a27dd93b4633eb8960b135

kornelski (approved):

boring/src/ssl/mod.rs
```diff
@@ -882,6 +882,52 @@ pub enum SslInfoCallbackValue {
...
+    /// Abort the handshake.
+    Error,
+
+    /// Continue with a full handshake.
+    ///
+    /// The peer supplied session ticket was not recognized.
+    ///
+    /// # Note
+    ///
+    /// This is a decryption specific status code.
```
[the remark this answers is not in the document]
toidiu: Updated in https://github.com/cloudflare/boring/pull/330/commits/9a6e16649e8097d6db0540ea097a6498710ff017 to indicate the status code is decryption specific when using submodule version of boring.

boring/src/ssl/test/session_resumption.rs
```diff
@@ -0,0 +1,272 @@
...
+    let server = server.build();
+
+    let mut client = server.client();
+    client
+        .ctx()
+        .set_session_cache_mode(SslSessionCacheMode::CLIENT);
+    client.ctx().set_new_session_callback(|_, session| {
+        NST_RECIEVED_COUNT.fetch_add(1, Ordering::SeqCst);
+        // The server sends multiple session tickets but we only care to retrieve one.
+        if SESSION_TICKET.get().is_none() {
```
rushilmehra, line 33: nit: since this is a OnceLock you can just always call `set` and ignore the error that gets returned the second time

boring/src/ssl/test/session_resumption.rs
```diff
@@ -0,0 +1,272 @@
...
+        }
+    });
+    let ssl_stream = client.connect();
+
+    assert!(!ssl_stream.ssl().session_reused());
+    assert!(SESSION_TICKET.get().is_some());
+    assert_eq!(NST_RECIEVED_COUNT.load(Ordering::SeqCst), 2);
+
+    // Retrieve the session ticket
+    let session_ticket = SslSession::from_der(SESSION_TICKET.get().unwrap()).unwrap();
```
rushilmehra, line 44: Instead of storing a `OnceLock<Vec<u8>>` and serializing / deserializing to and from der format, we can just store a `OnceLock<SslSession>`

boring/src/ssl/test/session_resumption.rs
toidiu: Yep makes sense. Updated inhttps://github.com/cloudflare/boring/pull/330/commits/cb41dbc1841b801a8fdf3628f4b9a0bb66b861a8

boring/src/ssl/test/session_resumption.rs
toidiu: Yep makes sense. Updated inhttps://github.com/cloudflare/boring/pull/330/commits/cb41dbc1841b801a8fdf3628f4b9a0bb66b861a8

boring/src/ssl/mod.rs
```diff
@@ -1185,6 +1232,43 @@ impl SslContextBuilder {
...
+    /// # Panics
+    ///
+    /// This method panics if this `Ssl` is associated with a RPK context.
+    #[corresponds(SSL_CTX_set_tlsext_ticket_key_cb)]
+    pub fn set_ticket_key_callback<F>(&mut self, callback: F)
+    where
+        F: Fn(
+                &SslRef,
+                &mut [u8; 16],
+                *mut u8,
```
rushilmehra, line 1254: The IV is constrained to a max length (https://github.com/google/boringssl/blob/main/ssl/ssl_session.cc#L331-L335) so instead of giving applications a raw pointer we should handle the unsafe conversion to a slice constrained to `ffi::EVP_MAX_IV_LENGTH` ourselves

boring/src/ssl/mod.rs
```diff
@@ -1185,6 +1232,43 @@ impl SslContextBuilder {
...
+                *mut ffi::EVP_CIPHER_CTX,
```
rushilmehra, line 1255: It's unfortunate that we have to pass these as raw pointers to users. How feasible is it to create safe bindings for these? If we do this then we'll have another task in our list for the 5.0 major bump

boring/src/ssl/mod.rs
toidiu: Yea good call. I wasn't sure if the length was dependent on the cipher.

boring/src/ssl/mod.rs
toidiu: > callback configures hmac_ctx with an HMAC digest and key, and configures ctx for encryption or decryption, based on the mode.

So based on the [docs](https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_CTX_set_tlsext_ticket_key_cb) the callback needs to configure the hmac and enc contexts.

For safe bindings I think we would need to expose a safe wrapper around [EVP_CipherInit_ex](https://commondatastorage.googleapis.com/chromium-boringssl-docs/cipher.h.html#EVP_CipherInit_ex), [EVP_DecryptInit_ex](https://commondatastorage.googleapis.com/chromium-boringssl-docs/cipher.h.html#EVP_DecryptInit_ex) and [HMAC_Init_ex](https://commondatastorage.googleapis.com/chromium-boringssl-docs/hmac.h.html#HMAC_Init_ex) and also make it extensible enough for various uses that customers might want.

From what I can tell its not trivial and would probably be a separate task than this PR since its complex enough.

boring/src/ssl/mod.rs
toidiu: > When encrypting a new ticket, encrypt will be one. It writes a public 16-byte key name to key_name and a fresh IV to iv.

Just to confirm, are we ok looking at BoringSSL implementation details to derive this API? The [docs](https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_CTX_set_tlsext_ticket_key_cb), quoted above only guarantee a 16 byte key name and a fresh IV (unspecified length).

boring/src/ssl/mod.rs
kornelski: To unblock this, we can make it `pub unsafe fn set_ticket_key_callback_unsafe`. The first `unsafe`, because we can't force the callback to do the correct initialization. The `_unsafe` suffix to leave room for a future `set_ticket_key_callback` that may have the right abstractions.

toidiu: > To unblock this, we can make it `pub unsafe fn set_ticket_key_callback_unsafe`. The first `unsafe`, because we can't force the callback to do the correct initialization. The `_unsafe` suffix to leave room for a future `set_ticket_key_callback` that may have the right abstractions.

Addressed in https://github.com/cloudflare/boring/pull/330/commits/91c95bb492d3d52942e17c1e834cb9261d17337e

boring/src/ssl/mod.rs
toidiu: > We can't force the callback initialize those fields. I think that makes it unsafe in Rust terms, because safe Rust code could leave them as-is, report success, and make boring read uninit bytes.

Addressed in https://github.com/cloudflare/boring/pull/330/commits/91c95bb492d3d52942e17c1e834cb9261d17337e.

Good catch. I went the `ptr::write(iv, [0; 16])` route.

I think we only want to initialized the values in `encrypt` mode since in `decrypt` mode boring will set the value for the callback to read. Lmk if I missed something.

boring/src/ssl/mod.rs
```diff
@@ -1180,6 +1227,50 @@ impl SslContextBuilder {
...
+    /// [`SSL_CTX_set_tlsext_ticket_key_cb`]: https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_CTX_set_tlsext_ticket_key_cb
+    #[corresponds(SSL_CTX_set_tlsext_ticket_key_cb)]
+    pub unsafe fn set_ticket_key_callback_unsafe<F>(&mut self, callback: F)
+    where
+        F: Fn(
+                &SslRef,
+                &mut [u8; 16],
+                &mut [u8; ffi::EVP_MAX_IV_LENGTH as usize],
+                &mut MaybeUninit<ffi::EVP_CIPHER_CTX>,
+                &mut MaybeUninit<ffi::HMAC_CTX>,
```
toidiu: > In both modes, ctx and hmac_ctx will already have been initialized with [EVP_CIPHER_CTX_init](https://commondatastorage.googleapis.com/chromium-boringssl-docs/cipher.h.html#EVP_CIPHER_CTX_init) and [HMAC_CTX_init](https://commondatastorage.googleapis.com/chromium-boringssl-docs/hmac.h.html#HMAC_CTX_init), respectively. callback configures hmac_ctx with an HMAC digest and key, and configures ctx for encryption or decryption, based on the mode.
>
> -- [Source](https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_CTX_set_tlsext_ticket_key_cb)

Based on the docs these should be initialized by Boring. The callback is only responsible for configuring the hmac/cipher_contexts not initializing them so I feel the `MaybeUninit` is going to lead to confusion.

boring/src/ssl/callbacks.rs
```diff
@@ -269,6 +270,67 @@ where
     }
 }
 
+unsafe fn to_uninit<'a, T: 'a>(ptr: *mut T) -> &'a mut MaybeUninit<T> {
+    assert!(!ptr.is_null());
+    unsafe { &mut *ptr.cast::<MaybeUninit<T>>() }
+}
```
toidiu, line 278: Confirming my understanding: panic on `assert` seems appropriate because otherwise we have UB.

boring/src/ssl/mod.rs
toidiu: > It's unfortunate that we have to pass these as raw pointers to users. How feasible is it to create safe bindings for these? If we do this then we'll have another task in our list for the 5.0 major bump

A safer interface added in this commit

https://github.com/cloudflare/boring/pull/330/commits/8c3bca21abb4527a9ebd53831126c783cd659843

boring/src/ssl/callbacks.rs
kornelski: References must never be null, so they can't be made from a null pointer.

This assert isn't strictly necessary as long as the function is always given non-null pointers. 

I thought it's better to be extra careful about this, I don't feel strongly about this choice. It could have been a `Result` (assuming null is a real possibility), or downgraded to `debug_assert` if we trust that we'll never get a null pointer in these functions.

boring/src/hmac.rs
```diff
@@ -0,0 +1,36 @@
...
+}
+
+impl HmacCtxRef {
+    /// Configures HmacCtx to use `md` as the hash function and `key` as the key.
+    ///
+    /// https://commondatastorage.googleapis.com/chromium-boringssl-docs/hmac.h.html#HMAC_Init_ex
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure HMAC_CTX has been initalized.
```
kornelski: I don't understand this safety requirement. It's not clear what "initialized" here refers to, since this is an `init` function doing initialization.

boring/src/symm.rs
```diff
@@ -68,6 +69,77 @@ pub enum Mode {
...
+}
+
+impl CipherCtxRef {
+    /// Configures CipherCtx for a fresh encryption operation using `cipher`.
+    ///
+    /// https://commondatastorage.googleapis.com/chromium-boringssl-docs/cipher.h.html#EVP_EncryptInit_ex
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure EVP_CIPHER_CTX has been initalized.
```
kornelski: again not clear what "initialized" is

boring/src/symm.rs
```diff
@@ -68,6 +69,77 @@ pub enum Mode {
...
+    ///
+    /// The caller is responsible for ensuring the length of `key` and `iv` are appropriate for the
```
kornelski: Only length of the `key` needs to be checked, right? The `iv` is an array of max len. 

Is there a cheap way to check key length against the `Cipher` to make it a safe function?

In `raw_ticket_key` the key is assumed to be 16 bytes long. Why is one place fixed, and another is flexible?

toidiu:
> It's nicer with safe wrapper types.

Yea, much nicer not to use the ffi functions, I was sold when I got rid of the `ret == 1` check.

> I've changed to use `ForeignTypeRef` to avoid `ManuallyDrop`.

Sweet. I actually had to debug a panic in tests before adding the ManuallyDrop so this seems like the correct choice.

boring/src/symm.rs
toidiu: Addressed in https://github.com/cloudflare/boring/pull/330/commits/2d5be20162b419ab86bf4c103fe701ec0c9f0bc7

> Only length of the key needs to be checked, right? The iv is an array of max len.
> Is there a cheap way to check key length against the Cipher to make it a safe function?

Yea good point. We can do a `key.len() != cipher.key_len()` and return early. But not sure if `ErrorStack::get()` is correct error to return here.

> In raw_ticket_key the key is assumed to be 16 bytes long. Why is one place fixed, and another is flexible?

raw_ticket_key is dealing with `key_name`, which is guaranteed to be 16 bytes ([docs](https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_CTX_set_tlsext_ticket_key_cb)).

Here, the application is calling `EVP_EncryptInit_ex` with the correct `key` and `iv`, the sizes of which depend on the chosen cipher.

I also expanded the tests to use the `iv` and `key_name` values from the callback.

boring/src/hmac.rs
toidiu: I was conflating documentation between this and session resumption. Removed and removed the `unsafe`.

boring/src/symm.rs
toidiu: I was conflating documentation between this and session resumption. Removed and removed the `unsafe`.

PR #331, opened 2025-03-11
boring-sys: Ignore patches when boringSSL is precompiled

cjpatton:
Internal users often have two builds for `boring`, one using a precompiled build of boringSSL and another built from source with patches applied. However the features that enable these builds are mutually exclusive. For example, the `"pq-experimental"` feature is required to build the source with all of the necessary codepoints for PQ key exchange, but if this feature is enabled and a precompiled boringSSL is provided, then the build will fail. This means users will have to also control their builds with mutually exclusive features.

An alternative is to *ignore* features that enable patches whenever a precompiled boringSSL is provided. This is a little different from the "assume patched" environment variable, which applies whenever we're building from source.

### DIFF
--- a/boring/src/ssl/mod.rs
+++ b/boring/src/ssl/mod.rs
@@ -718,15 +718,15 @@ impl SslCurve {
     pub const X25519_KYBER768_DRAFT00: SslCurve =
         SslCurve(ffi::SSL_CURVE_X25519_KYBER768_DRAFT00 as _);
 
-    #[cfg(feature = "pq-experimental")]
+    #[cfg(all(not(feature = "fips"), feature = "pq-experimental"))]
     pub const X25519_KYBER768_DRAFT00_OLD: SslCurve =
         SslCurve(ffi::SSL_CURVE_X25519_KYBER768_DRAFT00_OLD as _);
 
-    #[cfg(feature = "pq-experimental")]
+    #[cfg(all(not(feature = "fips"), feature = "pq-experimental"))]
     pub const X25519_KYBER512_DRAFT00: SslCurve =
         SslCurve(ffi::SSL_CURVE_X25519_KYBER512_DRAFT00 as _);
 
-    #[cfg(feature = "pq-experimental")]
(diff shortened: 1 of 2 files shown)


boring/src/ssl/mod.rs
```diff
@@ -718,14 +718,16 @@ impl SslCurve {
     pub const X25519_KYBER768_DRAFT00: SslCurve =
         SslCurve(ffi::SSL_CURVE_X25519_KYBER768_DRAFT00 as _);
 
-    #[cfg(feature = "pq-experimental")]
+    #[cfg(all(not(feature = "fips"), feature = "pq-experimental"))]
     pub const X25519_KYBER768_DRAFT00_OLD: SslCurve =
         SslCurve(ffi::SSL_CURVE_X25519_KYBER768_DRAFT00_OLD as _);
 
+    #[cfg(all(not(feature = "fips"), feature = "pq-experimental"))]
     #[cfg(feature = "pq-experimental")]
```
cjpatton, line 726: Redundant. LIkewise below.

boring/src/ssl/mod.rs
```diff
@@ -718,15 +718,15 @@ impl SslCurve {
     pub const X25519_KYBER768_DRAFT00: SslCurve =
         SslCurve(ffi::SSL_CURVE_X25519_KYBER768_DRAFT00 as _);
 
-    #[cfg(feature = "pq-experimental")]
+    #[cfg(all(not(feature = "fips"), feature = "pq-experimental"))]
```
cjpatton, line 721: `not(feature = "fips")` prevents us from building this code whenever `"fips"` is enabled. This is necessary because this API is too new for older boringSSL builds.

We could weaken this constraint to `not(feature = "fips-compat")`, but this is somewhat of a broader change than what this PR is trying to do.

Note: this is not a breaking change because `"fips"` and `"pq-experimental"` were previously mutually exclusive.

kornelski (approved):

cjpatton:
Rebased, plus I modified the warning to make it visible during the build.

Issue #196, opened 2023-12-01
Boring 5 Roadmap


nox:
Now that boring 4 is pretty old (I'm just joking), let's keep track of what we should do for boring 5:

* #194
* Rename `replace_ex_data` to `set_ex_data`, as keeping the leaking versions is a footgun
* Rename `Ssl::new_from_ref` to `Ssl::new`
* Rename `X509Builder::append_extension2` to `X509Builder::append_extension`
* Make `boring::ssl::SslStream::new` infallible, its current return type is `Result<_, ErrorStack>` but it returns `Ok(_)` always.
* Remove obsolete X509CheckFlags values
* Decouple "fips-compat" from "fips": https://github.com/cloudflare/boring/pull/333
* https://github.com/cloudflare/boring/issues/362
* https://github.com/cloudflare/boring/issues/363

howardjohn:
Any alternative (or addition, I suppose) to hyper 1.0 support could be https://github.com/cloudflare/boring/issues/165, which lets the rest of the rustls ecosystem handle things like tokio, hyper, etc integration

nox:
Added bullet point about `Clone` for `X509Store`.

kornelski:
We could unify interface for keys to match https://github.com/cloudflare/boring/pull/364

nox:
And we could probably remove `SslCipher` altogether and keep only `&'static SslCipherRef`.

PR #332, opened 2025-03-12
Add feature "fips-no-compat"

cjpatton:
As of boringSSL commit a430310d6563c0734ddafca7731570dfb683dc19, we no
longer need to make exceptions for the types of BufLen, ProtosLen, and
ValueLen, which means the "fips-compat" feature is no longer needed for
"fips" users.

Currently "fips" implies "fips-compat". To allow users to upgrade
without breaking API compatibility with boring version 4, add a new
feature, "fips-no-compat", that does not imply "fips-compat".

In boring 5, we should remove "fips-no-compat" and decouple
"fips-compat" from "fips".

### DIFF
--- a/boring/Cargo.toml
+++ b/boring/Cargo.toml
@@ -19,8 +19,18 @@ rustdoc-args = ["--cfg", "docsrs"]
 [features]
 # Controlling the build
 
-# Use a FIPS-validated version of boringssl.
+# Use a FIPS-validated version of BoringSSL. This feature sets "fips-compat".
 fips = ["fips-compat", "boring-sys/fips"]
+
+# Use a FIPS build of BoringSSL, but don't set "fips-compat".
+#
+# As of boringSSL commit a430310d6563c0734ddafca7731570dfb683dc19, we no longer
+# need to make exceptions for the types of BufLen, ProtosLen, and ValueLen,
+# which means the "fips-compat" feature is no longer needed.
+#
+# TODO(cjpatton) Delete this feature and modify "fips" so that it doesn't imply
+# "fips-compat".
+fips-no-compat = ["boring-sys/fips"]
 
 # Build with compatibility for the BoringSSL FIPS version, without enabling the
 # `fips` feature itself (useful e.g. if `fips-link-precompiled` is used with an
--- a/boring-sys/build/main.rs
+++ b/boring-sys/build/main.rs
@@ -760,7 +760,7 @@ fn main() {
         "des.h",
         "dtls1.h",
         "hkdf.h",
-        #[cfg(not(feature = "fips"))]
+        #[cfg(not(any(feature = "fips", feature = "fips-no-compat")))]
         "hpke.h",
         "hmac.h",
         "hrss.h",
--- a/boring/src/fips.rs
+++ b/boring/src/fips.rs
@@ -14,8 +14,16 @@ pub fn enabled() -> bool {
 
 #[test]
 fn is_enabled() {
-    #[cfg(any(feature = "fips", feature = "fips-link-precompiled"))]
+    #[cfg(any(
+        feature = "fips",
+        feature = "fips-no-compat",
+        feature = "fips-link-precompiled"
+    ))]
     assert!(enabled());
-    #[cfg(not(any(feature = "fips", feature = "fips-link-precompiled")))]
+    #[cfg(not(any(
+        feature = "fips",
+        feature = "fips-no-compat",
+        feature = "fips-link-precompiled"
+    )))]
     assert!(!enabled());
 }
--- a/boring/src/lib.rs
+++ b/boring/src/lib.rs
@@ -128,7 +128,7 @@ pub mod error;
 pub mod ex_data;
 pub mod fips;
 pub mod hash;
-#[cfg(not(feature = "fips"))]
+#[cfg(not(any(feature = "fips", feature = "fips-no-compat")))]
 pub mod hpke;
 pub mod memcmp;
 pub mod nid;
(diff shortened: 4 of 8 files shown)


cjpatton:
This breaks compatibility with users who need recent versions of boring (> 4.15.0) but older versions of boringSSL. An alternative would be to decouple "fips" and "fips-compat" so that "fips" doesn't imply "fips-compat". However, these users would need to update to include both features.

rushilmehra:
FYI this constitutes a breaking change because users that were previously getting `fips-compat` via the `fips` feature will now have to opt into `fips-compat` explicitly

rushilmehra:
I wonder if it's better to introduce a new feature (`fips-new` or something) that only enables `boring-sys/fips`, then we can clean this up when we want to cut 5.0

rushilmehra:
> > I wonder if it's better to introduce a new feature (`fips-new` or something) that only enables `boring-sys/fips`, then we can clean this up when we want to cut 5.0
> 
> (Perhaps `fips-<commit>` where `<commit>` is a boringSSL commit we know the bindings are compatible with.) I don't love this solution, as it seems like it would make our compiler checks more complex, e.g., `any(feature = "fips", feature = "fips-new")` instead of `fips(feature = "fips")`.
> 
> What's the timeline for 5.0?

```
❯ rg 'feature = "fips"'
boring/src/ssl/mod.rs
107:#[cfg(not(feature = "fips"))]
115:#[cfg(not(feature = "fips"))]
717:    #[cfg(not(feature = "fips"))]
762:            #[cfg(not(feature = "fips"))]
2013:    #[cfg(not(feature = "fips"))]
2270:    #[cfg(not(feature = "fips"))]
3692:    #[cfg(not(feature = "fips"))]
3711:    #[cfg(not(feature = "fips"))]
3733:    #[cfg(not(feature = "fips"))]
3750:    #[cfg(not(feature = "fips"))]
3757:    #[cfg(not(feature = "fips"))]

boring/src/ssl/test/mod.rs
24:#[cfg(not(feature = "fips"))]
30:#[cfg(not(feature = "fips"))]
993:#[cfg(not(feature = "fips"))]

boring/src/fips.rs
17:    #[cfg(any(feature = "fips", feature = "fips-link-precompiled"))]
19:    #[cfg(not(any(feature = "fips", feature = "fips-link-precompiled")))]

boring/src/lib.rs
131:#[cfg(not(feature = "fips"))]

boring-sys/build/main.rs
763:        #[cfg(not(feature = "fips"))]
```
Luckily there's not too many instances, so it's not the worst thing in the world. Cutting 5.0 is something we can do whenever we'd like, but it would take a lot more effort to update the entire world internally and we'd probably want to take a pass at landing a bunch of other breaking changes in the pipeline

boring/Cargo.toml
```diff
@@ -19,9 +19,19 @@ rustdoc-args = ["--cfg", "docsrs"]
...
+# Use a FIPS-validated version of BoringSSL. This feature sets "fips-compat".
 fips = ["fips-compat", "boring-sys/fips"]
 
+# Use a FIPS build of BoringSSL, but don't set "fips-compat".
+#
+# As of boringSSL commit a430310d6563c0734ddafca7731570dfb683dc19, we no longer
+# need to make exceptions for the types of BufLen, ProtosLen, and ValueLen,
+# which means the "fips-compat" feature is no longer needed.
+#
+# TODO(cjpatton) Delete this feature and modify "fips" so that it doesn't imply
```
rushilmehra, line 31: Lets a bullet to https://github.com/cloudflare/boring/issues/196

rushilmehra:
Not to bikeshed, but wdyt about `fips-modern`? Not a blocker at all

cjpatton:
> Not to bikeshed, but wdyt about `fips-modern`? Not a blocker at all

I wanted the semantics to be crystal clear. I also think an uglier name like `fips-no-compat` will help remind us to remove this.

boring/Cargo.toml
cjpatton: Done, and threw in a PR: https://github.com/cloudflare/boring/pull/333

cjpatton:
> I wonder if it's better to introduce a new feature (`fips-new` or something) that only enables `boring-sys/fips`, then we can clean this up when we want to cut 5.0

Done.

boring/Cargo.toml
```diff
@@ -19,9 +19,19 @@ rustdoc-args = ["--cfg", "docsrs"]
...
+# "fips-compat".
+fips-no-compat = ["boring-sys/fips"]
```
cjpatton, line 33: Note: None of the testes in CI exercise this feaature, though I've validated it in the internal oxy repo.

I could add tests, but we'd also need to bump `boring-sys/deps/boringssl-fips`, which is used to validate older builds. We could add another boringSSL submodule, but that would be a little annoying.

cjpatton:
Ah darn, I just realized we need to implement the same pattern for the hyper-boring and tokio-boring crates :( Back to draft while I work on this.

boring/Cargo.toml
rushilmehra: You could make a test that pulls a recent version of bssl to build bindings with, but that seems a bit arduous. The change isn't particularly scary, so I wouldn't say it's strictly necessary. If things pass internally then that's good. When we cut 5.0 we'll make sure CI is up to date

rushilmehra:
> Ah darn, I just realized we need to implement the same pattern for the hyper-boring and tokio-boring crates :( Back to draft while I work on this.

Shouldn't be too bad - the `fips` features in the other crates just shim to `boring/fips` and `boring-sys/fips`. So we just need to mirror that for `fips-no-compat`

rushilmehra (approved):

PR #336, opened 2025-03-18
Remove "fips-no-compat", decouple "fips-compat" from "fips"

cjpatton:
Partially addresses https://github.com/cloudflare/boring/issues/196.

NOTE: This is an API breaking change, so we're waiting to merge until we're ready to cut the next major version.

Modify the "fips" feature so that it no longer implies "fips-compat". The latter is no longer needed for recent builds of boringSSL; users who need older builds will need to enable "fips-compat" explicitly.

Also, remove the "fipps-no-compat" feature, as it's now equivalent to "fips".


cjpatton:
Closed in favor of #338.

PR #338, opened 2025-03-20
Add fips-precompiled feature to support newer versions of FIPS

rushilmehra:
Newer versions of FIPS don't need any special casing in our bindings,
unlike the submoduled boringssl-fips. In addition, many users currently
use FIPS by precompiling BoringSSL with the proper build tools and
passing that in to the bindings.

Until we adopt the Update Stream pattern for FIPS, there are two main
use cases:

1. Passing an unmodified, precompiled FIPS validated version of
   boringssl (fips-precompiled)

2. Passing a custom source directory of boringssl meant to be linked
   with a FIPS validated bcm.o. This is mainly useful if you carry
   custom patches but still want to use a FIPS validated BoringCrypto.
   (fips-link-precompiled)

This commit introduces the `fips-precompiled` feature and removes the
`fips-no-compat` feature.

### DIFF
--- a/boring-sys/Cargo.toml
+++ b/boring-sys/Cargo.toml
@@ -57,8 +57,18 @@ features = ["rpk", "pq-experimental", "underscore-wildcards"]
 rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
-# Use a FIPS-validated version of boringssl.
+# Compile boringssl using the FIPS build flag if building boringssl from
+# scratch.
+#
+# See
+# https://boringssl.googlesource.com/boringssl/+/master/crypto/fipsmodule/FIPS.md
+# for instructions and more details on the boringssl FIPS flag.
 fips = []
+
+# Use a precompiled FIPS-validated version of BoringSSL. Meant to be used with
+# FIPS-20230428 or newer. Users must set `BORING_BSSL_FIPS_PATH` to use this
+# feature, or else the build will fail.
+fips-precompiled = []
 
 # Link with precompiled FIPS-validated `bcm.o` module.
 fips-link-precompiled = []
--- a/boring-sys/build/config.rs
+++ b/boring-sys/build/config.rs
@@ -16,6 +16,7 @@ pub(crate) struct Config {
 
 pub(crate) struct Features {
     pub(crate) fips: bool,
+    pub(crate) fips_precompiled: bool,
     pub(crate) fips_link_precompiled: bool,
     pub(crate) pq_experimental: bool,
     pub(crate) rpk: bool,
@@ -47,11 +48,7 @@ impl Config {
         let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap();
 
         let features = Features::from_env();
-        let env = Env::from_env(
-            &host,
-            &target,
-            features.fips || features.fips_link_precompiled,
-        );
+        let env = Env::from_env(&host, &target, features.is_fips_like());
 
         let mut is_bazel = false;
         if let Some(src_path) = &env.source_path {
@@ -80,6 +77,10 @@ impl Config {
             panic!("`fips` and `rpk` features are mutually exclusive");
         }
 
+        if self.features.fips_precompiled && self.features.rpk {
+            panic!("`fips-precompiled` and `rpk` features are mutually exclusive");
+        }
+
         let is_precompiled_native_lib = self.env.path.is_some();
         let is_external_native_lib_source =
             !is_precompiled_native_lib && self.env.source_path.is_none();
@@ -103,8 +104,14 @@ impl Config {
             );
         }
 
+        // todo(rmehra): should this even be a restriction? why not let people link a custom bcm.o?
+        // precompiled boringssl will include libcrypto.a
         if is_precompiled_native_lib && self.features.fips_link_precompiled {
             panic!("precompiled BoringSSL was provided, so FIPS configuration can't be applied");
+        }
+
+        if !is_precompiled_native_lib && self.features.fips_precompiled {
+            panic!("`fips-precompiled` feature requires `BORING_BSSL_FIPS_PATH` to be set");
         }
     }
 }
@@ -112,6 +119,7 @@ impl Config {
 impl Features {
     fn from_env() -> Self {
         let fips = env::var_os("CARGO_FEATURE_FIPS").is_some();
+        let fips_precompiled = env::var_os("CARGO_FEATURE_FIPS_PRECOMPILED").is_some();
         let fips_link_precompiled = env::var_os("CARGO_FEATURE_FIPS_LINK_PRECOMPILED").is_some();
         let pq_experimental = env::var_os("CARGO_FEATURE_PQ_EXPERIMENTAL").is_some();
         let rpk = env::var_os("CARGO_FEATURE_RPK").is_some();
@@ -119,11 +127,16 @@ impl Features {
 
         Self {
             fips,
+            fips_precompiled,
             fips_link_precompiled,
             pq_experimental,
             rpk,
             underscore_wildcards,
         }
+    }
+
+    pub(crate) fn is_fips_like(&self) -> bool {
+        self.fips || self.fips_precompiled || self.fips_link_precompiled
     }
 }
 
@@ -138,6 +151,7 @@ impl Env {
         let target_var = |name: &str| {
             let kind = if host == target { "HOST" } else { "TARGET" };
 
+            // TODO(rmehra): look for just `name` first, as most people just set that
             var(&format!("{}_{}", name, target))
                 .or_else(|| var(&format!("{}_{}", name, target_with_underscores)))
                 .or_else(|| var(&format!("{}_{}", kind, name)))
--- a/boring/Cargo.toml
+++ b/boring/Cargo.toml
@@ -19,23 +19,26 @@ rustdoc-args = ["--cfg", "docsrs"]
 [features]
 # Controlling the build
 
-# Use a FIPS-validated version of BoringSSL. This feature sets "fips-compat".
+# NOTE: This feature is deprecated. It is needed for the submoduled
+# boringssl-fips, which is extremely old and requires modifications to the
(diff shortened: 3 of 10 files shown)


boring-sys/Cargo.toml
```diff
@@ -60,6 +60,9 @@ rustdoc-args = ["--cfg", "docsrs"]
 # Use a FIPS-validated version of boringssl.
 fips = []
 
+# Use a precompiled, FIPS validated boringssl.
+fips-precompiled = []
+
 # Link with precompiled FIPS-validated `bcm.o` module.
 fips-link-precompiled = []
```
cjpatton, line 67: The descriptions for each of these flags all sound basically the same. I would be more specific about what the new flag actually does, in particular how it differs from "fips".

Also, "fips-precompiled" is very similar to "fips-link-precompiled".

rushilmehra: I added more detailed documentation in https://github.com/cloudflare/boring/pull/338/files#diff-902ee088de98c256522122e618b15e1fb66c1e57f7e30a0d108cefcb1af7df19R22 since it's the most popular entrypoint to the crate. I can copy documentation over to `boring-sys/Cargo.toml` if you'd like

rushilmehra: Copied the relevant documentation over

boring-sys/Cargo.toml
```diff
@@ -57,9 +57,15 @@ features = ["rpk", "pq-experimental", "underscore-wildcards"]
 rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
-# Use a FIPS-validated version of boringssl.
+# Compile boringssl using the FIPS build flag if building boringssl from
+# scratch.
```
cjpatton, line 61: What is the "FIPS build flag"? Are you referring to a way to build boringSSL itself? If so, perhaps add a reference to the documentation.

boring-sys/build/config.rs
```diff
@@ -80,6 +77,10 @@ impl Config {
             panic!("`fips` and `rpk` features are mutually exclusive");
         }
 
+        if self.features.fips_precompiled && self.features.rpk {
+            panic!("`fips-precompiled` and `rpk` features are mutually exclusive");
```
cjpatton, line 81: Why are these mutually exclusive?

boring/Cargo.toml
```diff
@@ -19,24 +19,27 @@ rustdoc-args = ["--cfg", "docsrs"]
...
 
-# Use a FIPS-validated version of BoringSSL. This feature sets "fips-compat".
-fips = ["fips-compat", "boring-sys/fips"]
-
-# Use a FIPS build of BoringSSL, but don't set "fips-compat".
+# NOTE: This feature is deprecated. It panders to the submoduled
+# boringssl-fips, which is extremely old and requires modifications to the
+# bindings, as some newer APIs don't exist and some function signatures have
+# changed. It is highly recommended to use `fips-precompiled` instead.
 #
```
cjpatton, line 26: Since we haven't actually published 4.16, I would argue that we could just yank the tag and delete `fips-no-compat` before we cut and publish 4.17.

boring/Cargo.toml
```diff
@@ -19,24 +19,27 @@ rustdoc-args = ["--cfg", "docsrs"]
...
-# need to make exceptions for the types of BufLen, ProtosLen, and ValueLen,
-# which means the "fips-compat" feature is no longer needed.
+# This feature sets `fips-compat` on behalf of the user to guarantee bindings
+# compatibility with the submoduled boringssl-fips.
 #
-# TODO(cjpatton) Delete this feature and modify "fips" so that it doesn't imply
-# "fips-compat".
-fips-no-compat = ["boring-sys/fips"]
+# Use a FIPS-validated version of BoringSSL.
+fips = ["fips-compat", "boring-sys/fips"]
```
cjpatton, line 31: Question: do we plan to modify/delete the `fips` feature as part of boring 5? I would be in favor of replacing `fips` with `fips-precompiled`, since this will allow internal users to migrate to the latest fips build without changing their feature set.

On the other hand, users of old fips builds will have to change their feature set, but there are less such users.

cjpatton (changes requested):
I'm on board with this plan! However the PR isn't ready to merge, as there are a couple of issues building against the latest boringssl fips build.

First, some of the bindings are unavailable:
```
error[E0425]: cannot find value `SSL_CURVE_X25519_KYBER768_DRAFT00` in crate `ffi`
    --> boring/src/ssl/mod.rs:719:23
     |
719  |         SslCurve(ffi::SSL_CURVE_X25519_KYBER768_DRAFT00 as _);
     |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: a constant with a similar name exists: `SSL_CURVE_X25519KYBER768DRAFT00`
     |
    ::: /cfsetup_build/boring/target/debug/build/boring-sys-187d757bde24256b/out/bindings.rs:4383:1
     |
4383 | pub const SSL_CURVE_X25519KYBER768DRAFT00: i32 = 25497;
     | ---------------------------------------------- similarly named constant `SSL_CURVE_X25519KYBER768DRAFT00` defined here

error[E0531]: cannot find unit struct, unit variant or constant `SSL_CURVE_X25519_KYBER768_DRAFT00` in crate `ffi`
    --> boring/src/ssl/mod.rs:763:18
     |
763  |             ffi::SSL_CURVE_X25519_KYBER768_DRAFT00 => Some(ffi::NID_X25519Kyber768Draft00),
     |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: a constant with a similar name exists: `SSL_CURVE_X25519KYBER768DRAFT00`
     |
    ::: /cfsetup_build/boring/target/debug/build/boring-sys-187d757bde24256b/out/bindings.rs:4383:1
     |
4383 | pub const SSL_CURVE_X25519KYBER768DRAFT00: i32 = 25497;
     | ---------------------------------------------- similarly named constant `SSL_CURVE_X25519KYBER768DRAFT00` defined here
```

I think we should just disable these APIs, i.e., via `#[cfg(not(any(feature = "fips", feature = "fips-precompiled")))]`.

Second, in my build environment, it looks like the linker is failing to find libcrypto:
```
chris@docker-desktop:/cfsetup_build/boring$ export BORING_BSSL_FIPS_PATH="/opt/boringssl-fips"
chris@docker-desktop:/cfsetup_build/boring$ cargo test --features fips-precompiled
   Compiling boring-sys v4.15.0 (/cfsetup_build/boring/boring-sys)
   Compiling boring v4.15.0 (/cfsetup_build/boring/boring)
error: could not find native static library `crypto`, perhaps an -L flag is missing?

error: could not compile `boring-sys` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
chris@docker-desktop:/cfsetup_build/boring$
```

Finally, the commit message should mention that we're deleting `fips-no-compat`.

boring/Cargo.toml
rushilmehra: I'm not sure what you mean, we didn't release 4.16 so we can just delete the code containing `fips-no-compat` and release 4.16

boring-sys/build/config.rs
rushilmehra: Because we don't ship our precompiled FIPS package with the RPK patch applied

boring-sys/Cargo.toml
rushilmehra: Yes it's a boringssl build flag, I added a link

boring/Cargo.toml
rushilmehra: Yeah ideally in 5.0 `fips` will become `fips-precompiled`

rushilmehra:
> Second, in my build environment, it looks like the linker is failing to find libcrypto:

I fixed the build issues
 
 > I think we should just disable these APIs, i.e., via `#[cfg(not(any(feature = "fips", feature = "fips-precompiled")))]`.

I've done this, but the real issue here is that the PQ patch in our internal FIPS precompiled package and the patch we ship in this repo have diverged. 

@bwesterb Do you have thoughts here? Do we want to continue maintaining the PQ patch in this repo, or can public users of the crate just rely on the PQ codepoints that are in boringssl HEAD?

boring/Cargo.toml
```diff
@@ -19,24 +19,27 @@ rustdoc-args = ["--cfg", "docsrs"]
 [features]
 # Controlling the build
 
-# Use a FIPS-validated version of BoringSSL. This feature sets "fips-compat".
-fips = ["fips-compat", "boring-sys/fips"]
-
-# Use a FIPS build of BoringSSL, but don't set "fips-compat".
+# NOTE: This feature is deprecated. It panders to the submoduled
```
cjpatton, line 22:
```suggestion
# NOTE: This feature is deprecated. It is needed for the submoduled
```

cjpatton (approved):
> > Second, in my build environment, it looks like the linker is failing to find libcrypto:
> 
> I fixed the build issues

Looks good to me now!

> > I think we should just disable these APIs, i.e., via `#[cfg(not(any(feature = "fips", feature = "fips-precompiled")))]`.
> 
> I've done this, but the real issue here is that the PQ patch in our internal FIPS precompiled package and the patch we ship in this repo have diverged.
> 
> @bwesterb Do you have thoughts here? Do we want to continue maintaining the PQ patch in this repo, or can public users of the crate just rely on the PQ codepoints that are in boringssl HEAD?

I think the main reason we still need `pq-experimental` is to support older codepoints, namely `X25519Kyber768Draft00` and `P256Kyber768Draft00`. I'd imagine we'd need to disable these slowly, starting with the client and then moving onto the server.

boring-sys/build/config.rs
rushilmehra: Yeah I agree - though right now we panic when the `fips` and `rpk` changes are enabled, and ideally we follow what you're saying consistently throughout the codebase. For consistency's sake, I think we can leave it as is, and in a follow-up PR I'll adopt the approach you're suggesting more generally

PR #339, opened 2025-03-22
feat(x509): Implement `Clone` for `X509Store`

0x676e67:
Some scenarios require repeatedly creating TLS connectors. We aim to reuse X509Store instead of recreating it each time. Implementing Clone could be a good solution.


nox:
Reverting this in https://github.com/cloudflare/boring/pull/353.

PR #345, opened 2025-05-01
add SslCurve::X25519_MLKEM768 constant

ehaydenr:


cjpatton (approved):

bwesterb:
Aside: it's better to use `set_curves_list`.

kornelski:
The constants are already there, so it seems sensible to me to keep adding them to keep the list consistent with the underlying library.

If `set_curves_list` is strongly preferred, maybe we could mark all of the constants as deprecated.

cjpatton:
> Why is `set_curves_list` strongly preferred? When calling `boring::ssl::Ssl::curve()`, the `SslCurve` constants allow for convenient comparison/equality checks.

From my perspective, the `set_curves_list()` has been the much more stable API overtime. With previous builds of boringSSL we've had to deal with naming changes that cause `set_curves()` API to not compile.

PR #346, opened 2025-05-02
Expose SSL_set1_groups to Efficiently Set Curves on SSL Session

Justin-Kwan:
**Description**
- Exposes BoringSSL method `SslRef::set_groups` -> `SSL_set1_groups` to directly set curve groups against an ongoing SSL session. This is necessary because in boringssl, calling `SSL_set_SSL_CTX` to switch the `SSL_CTX` does not update the supported group list (supported_group_list) on the existing SSL session object. We must set the curve list directly on the session after the SSL object is created but before the handshake begins.
b377876c-b3eb-492f-b392-b18746ba605a

- `SSL_set1_groups` is more efficient than the currently exposed `SslRef::set_curves_list` -> `SSL_set1_groups_list` which uses more complicated parsing logic to [parse colon delimited](https://github.com/google/boringssl/blob/bf4cf6938a77f1aca83ef529dce96681efd1e6c5/ssl/ssl_lib.cc#L1861) groups in string
113e324c-a2a1-489f-b65c-6300f1366c8a

- Exposes BoringSSL method `SslRef::set_options` -> `SSL_set_options` to directly set options against an ongoing SSL session (ex. enable `boring::ssl::SslOptions::CIPHER_SERVER_PREFERENCE` for a specific SNI only)

### DIFF
--- a/boring/src/ssl/mod.rs
+++ b/boring/src/ssl/mod.rs
@@ -767,7 +767,7 @@ impl SslCurve {
     // underlying boringssl version is upgraded, this should be removed in favor of the new
     // SSL_CTX_set1_group_ids API.
     #[allow(dead_code)]
-    fn nid(&self) -> Option<c_int> {
+    pub fn nid(&self) -> Option<c_int> {
         match self.0 {
             ffi::SSL_CURVE_SECP224R1 => Some(ffi::NID_secp224r1),
             ffi::SSL_CURVE_SECP256R1 => Some(ffi::NID_X9_62_prime256v1),
@@ -2813,6 +2813,25 @@ impl SslRef {
         unsafe { ffi::SSL_get_rbio(self.as_ptr()) }
     }
 
+    /// Sets the options used by the ongoing session, returning the old set.
+    ///
+    /// # Note
+    ///
+    /// This *enables* the specified options, but does not disable unspecified options. Use
+    /// `clear_options` for that.
+    #[corresponds(SSL_set_options)]
+    pub fn set_options(&mut self, option: SslOptions) -> SslOptions {
+        let bits = unsafe { ffi::SSL_set_options(self.as_ptr(), option.bits()) };
+        SslOptions::from_bits_retain(bits)
+    }
+
+    /// Clears the options used by the ongoing session, returning the old set.
+    #[corresponds(SSL_clear_options)]
+    pub fn clear_options(&mut self, option: SslOptions) -> SslOptions {
+        let bits = unsafe { ffi::SSL_clear_options(self.as_ptr(), option.bits()) };
+        SslOptions::from_bits_retain(bits)
+    }
+
     #[corresponds(SSL_set1_curves_list)]
     pub fn set_curves_list(&mut self, curves: &str) -> Result<(), ErrorStack> {
         let curves = CString::new(curves).unwrap();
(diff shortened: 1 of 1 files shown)


cjpatton:
Does https://docs.rs/boring/latest/boring/ssl/struct.SslContextBuilder.html#method.set_curves work for you?

Justin-Kwan:
Hi @cjpatton! Unfortunately, [SslContextBuilder::set_curves](https://docs.rs/boring/latest/boring/ssl/struct.SslContextBuilder.html#method.set_curves) only applies the curve list to the SSL_CTX object, boringssl does not propagate those curves to new SSL session objects created from that context.

bwesterb:
We're talking about deprecating set_curves in favour of set_curves_list. Does that cause problems for you?

Justin-Kwan:
Hi @bwesterb. Are you referring to the `[SSL_set1_curves_list](https://www.openssl.org/docs/manmaster/man3/SSL_set1_curves_list.html)` onto the SSL session itself?

Ideally, we'd like to use the `set_curves` method directly, to avoid any colon delimited string parsing overhead upon each new SSL session.

https://github.com/google/boringssl/blob/bf4cf6938a77f1aca83ef529dce96681efd1e6c5/ssl/ssl_lib.cc#L1861

boring/src/ssl/mod.rs
```diff
@@ -767,7 +767,7 @@ impl SslCurve {
     // underlying boringssl version is upgraded, this should be removed in favor of the new
     // SSL_CTX_set1_group_ids API.
     #[allow(dead_code)]
-    fn nid(&self) -> Option<c_int> {
+    pub fn nid(&self) -> Option<c_int> {
```
kornelski, line 770: We don't expose `c_int` in the public API. Please wrap it in a newtype, like other typed constants, e.g.

```rust

#[derive(Debug, Copy, Clone, PartialEq, Eq)]
pub struct SniError(c_int);

impl SniError {
    pub const ALERT_FATAL: SniError = SniError(ffi::SSL_TLSEXT_ERR_ALERT_FATAL);
    pub const ALERT_WARNING: SniError = SniError(ffi::SSL_TLSEXT_ERR_ALERT_WARNING);
}
```

boring/src/ssl/mod.rs
```diff
@@ -2826,6 +2845,20 @@ impl SslRef {
         }
     }
 
+    /// Sets the ongoing session's supported groups by their named identifiers
+    /// (formerly referred to as curves).
+    #[corresponds(SSL_set1_groups)]
+    pub fn set_group_nids(&mut self, group_nids: &[i32]) -> Result<(), ErrorStack> {
```
kornelski, line 2851: `&[i32]` should use the newtype (BTW, `i32` isn't guaranteed to be `c_int`)

If you use `#[repr(transparent)]` on the newtype, then it will be safe to cast the pointer to the slice, without copying.

kornelski (changes requested):
Looks okay, except I'd prefer a newtype instead of `c_int` in the API.

kornelski (approved):

Justin-Kwan:
Thanks for the feedback and review @kornelski!

boring/src/ssl/mod.rs
```diff
@@ -2825,6 +2853,20 @@ impl SslRef {
         }
     }
 
+    /// Sets the ongoing session's supported groups by their named identifiers
+    /// (formerly referred to as curves).
+    #[corresponds(SSL_set1_groups)]
+    pub fn set_group_nids(&mut self, group_nids: &[SslCurveNid]) -> Result<(), ErrorStack> {
+        unsafe {
+            cvt_0i(ffi::SSL_set1_groups(
```
kornelski: You need to change it to the older `SSL_CTX_set1_curves` for compatibility with our FIPS builds. `SSL_set1_groups` just calls `SSL_CTX_set1_curves`, so it's the same thing.

kornelski:
```suggestion
            cvt_0i(ffi::SSL_set1_curves(
```

Justin-Kwan:
@kornelski Thank you for the help, happy to see build is passing!

PR #349, opened 2025-05-19
Add `X509_STORE_CTX_get0_cert` interface

semaj-cf:
This method reliably retrieves the certificate the `X509_STORE_CTX` is verifying, unlike `X509_STORE_CTX_get_current_cert`, which may return the "problematic" cert when verification fails.


semaj-cf:
@bwesterb @nox

bwesterb (approved):

PR #359, opened 2025-06-02
Improve X509StoreContext initialization API

nox:
We introduce `X509StoreContextRef::init_without_cleanup`.

Function `X509_STORE_CTX_init` requires its arguments to outlive the store context, which is why `X509StoreContextRef::init` takes a closure which is guaranteed to be followed by a call to `X509_STORE_CTX_cleanup`.

We can avoid that by using the store context's ex data to store and own the arguments that were passed to `X509_STORE_CTX_init`.

### DIFF
--- a/boring/src/lib.rs
+++ b/boring/src/lib.rs
@@ -108,6 +108,8 @@ extern crate libc;
 #[cfg(test)]
 extern crate hex;
 
+use std::ffi::{c_long, c_void};
+
 #[doc(inline)]
 pub use crate::ffi::init;
 
@@ -191,5 +193,18 @@ fn cvt_n(r: c_int) -> Result<c_int, ErrorStack> {
         Err(ErrorStack::get())
     } else {
         Ok(r)
+    }
+}
+
+unsafe extern "C" fn free_data_box<T>(
+    _parent: *mut c_void,
+    ptr: *mut c_void,
+    _ad: *mut ffi::CRYPTO_EX_DATA,
+    _idx: c_int,
+    _argl: c_long,
+    _argp: *mut c_void,
+) {
+    if !ptr.is_null() {
+        drop(Box::<T>::from_raw(ptr as *mut T));
     }
 }
--- a/boring/src/x509/mod.rs
+++ b/boring/src/x509/mod.rs
@@ -131,8 +131,6 @@ impl X509StoreContextRef {
     ///
     /// This can be used to provide data to callbacks registered with the context. Use the
     /// `Ssl::new_ex_index` method to create an `Index`.
-    ///
-    /// The previous value, if any, will be returned.
     #[corresponds(X509_STORE_CTX_set_ex_data)]
     pub fn set_ex_data<T>(&mut self, index: Index<X509StoreContext, T>, data: T) {
         if let Some(old) = self.ex_data_mut(index) {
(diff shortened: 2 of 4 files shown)


boring/src/lib.rs
```diff
@@ -193,3 +195,16 @@ fn cvt_n(r: c_int) -> Result<c_int, ErrorStack> {
...
+unsafe extern "C" fn free_data_box<T>(
+    _parent: *mut c_void,
+    ptr: *mut c_void,
+    _ad: *mut ffi::CRYPTO_EX_DATA,
+    _idx: c_int,
+    _argl: c_long,
+    _argp: *mut c_void,
+) {
+    if !ptr.is_null() {
+        drop(Box::<T>::from_raw(ptr as *mut T));
```
kornelski, line 208: There's `ptr.cast::<T>()`

boring/src/x509/mod.rs
```diff
@@ -142,6 +206,38 @@ impl X509StoreContextRef {
...
+        cert_chain: Stack<X509>,
+    ) -> Result<(), ErrorStack> {
+        unsafe {
+            if let Err(e) = cvt(ffi::X509_STORE_CTX_init(
+                self.as_ptr(),
+                trust.as_ptr(),
+                cert.as_ptr(),
+                cert_chain.as_ptr(),
+            )) {
+                ffi::X509_STORE_CTX_cleanup(self.as_ptr());
```
kornelski, line 228: the method is called "without cleanup", and still may call cleanup?

boring/src/x509/mod.rs
```diff
@@ -142,6 +206,38 @@ impl X509StoreContextRef {
...
+        }
+
+        self.set_ex_data(*STORE_INDEX, trust);
+        self.set_ex_data(*CERT_INDEX, cert);
+        self.set_ex_data(*CERT_CHAIN_INDEX, cert_chain);
+
+        Ok(())
+    }
+
+    /// Sets the X509 verification configuration.
```
kornelski: copypaste error

boring/src/x509/mod.rs
nox: Yes on the error path. `X509_STORE_CTX_init` also starts with a `X509_STORE_CTX_cleanup` call.

boring/src/x509/mod.rs
nox: Is it? `verify_param_mut` calls it the X509 verification configuration too.

boring/src/x509/mod.rs
nox: I'm open to suggestions for a better name btw.

boring/src/x509/mod.rs
kornelski: I mean docs for verify_param are wrong. The method calls `get`, but the doc comment and `corresponds` say it's a setter.

boring/src/x509/mod.rs
kornelski: I'm not sure how to name the bundle of trust + cert + cert_chain. Context? It's sort of context, but not exactly X509StoreContextRef. Let's say `context_data`.

Then to make it clear it keeps them, `initialized_with_context_data` or `set_context_data`? Or take `self` and `with_context_data`.

nox: What about `reset` or `reset_with_context_data`?

kornelski: Yeah, that works.

PR #364, opened 2025-06-03
Expose PKey::raw_{private,public}_key

jrose-signal:
Had occasion to want these, hopefully relatively uncontroversial. I did see that there are two different ways of adapting the "measure once, output once" APIs: SslSession methods follow this approach, using an `&mut [u8]` argument and special-casing the empty case (though apparently that's done within BoringSSL itself); `EcPointRef::to_bytes`, by contrast, just calls the underlying BoringSSL API twice so it can return a `Vec<u8>`. Since the caller must already know something about the kind of key to expect success, I figured it was worth providing the more flexible slice-based API.

### DIFF
--- a/boring/src/pkey.rs
+++ b/boring/src/pkey.rs
@@ -223,23 +223,35 @@ where
         unsafe { ffi::EVP_PKEY_cmp(self.as_ptr(), other.as_ptr()) == 1 }
     }
 
+    /// Returns the length of the "raw" form of the public key. Only supported for certain key types.
+    ///
+    /// Returns the used portion of `out`.
+    #[corresponds(EVP_PKEY_get_raw_public_key)]
+    pub fn raw_public_key_len(&self) -> Result<usize, ErrorStack> {
+        unsafe {
+            let mut size = 0;
+            _ = cvt_0i(ffi::EVP_PKEY_get_raw_public_key(
+                self.as_ptr(),
+                std::ptr::null_mut(),
+                &mut size,
+            ))?;
+            Ok(size)
+        }
+    }
+
     /// Outputs a copy of the "raw" form of the public key. Only supported for certain key types.
     ///
-    /// Returns the number of bytes written, or the size of the raw form if `out` is empty.
+    /// Returns the used portion of `out`.
     #[corresponds(EVP_PKEY_get_raw_public_key)]
-    pub fn raw_public_key(&self, out: &mut [u8]) -> Result<usize, ErrorStack> {
+    pub fn raw_public_key<'a>(&self, out: &'a mut [u8]) -> Result<&'a [u8], ErrorStack> {
         unsafe {
             let mut size = out.len();
             _ = cvt_0i(ffi::EVP_PKEY_get_raw_public_key(
                 self.as_ptr(),
-                if size == 0 {
-                    std::ptr::null_mut()
-                } else {
-                    out.as_mut_ptr()
-                },
+                out.as_mut_ptr(),
                 &mut size,
             ))?;
-            Ok(size)
+            Ok(&out[..size])
         }
     }
 }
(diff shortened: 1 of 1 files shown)


boring/src/pkey.rs
```diff
@@ -260,6 +280,26 @@ where
         private_key_to_der_pkcs8_passphrase,
         ffi::i2d_PKCS8PrivateKey_bio
     }
+
+    /// Outputs a copy of the "raw" form of the private key. Only supported for certain key types.
+    ///
+    /// Returns the number of bytes written, or the size of the raw form if `out` is empty.
+    #[corresponds(EVP_PKEY_get_raw_private_key)]
+    pub fn raw_private_key(&self, out: &mut [u8]) -> Result<usize, ErrorStack> {
```
nox: I would rather have two methods, `raw_private_key` and `raw_private_key_len`, rather than piggyback `raw_private_key` to do a different thing when `out` is empty.

boring/src/pkey.rs
```diff
@@ -222,6 +222,26 @@ where
     {
         unsafe { ffi::EVP_PKEY_cmp(self.as_ptr(), other.as_ptr()) == 1 }
     }
+
+    /// Outputs a copy of the "raw" form of the public key. Only supported for certain key types.
+    ///
+    /// Returns the number of bytes written, or the size of the raw form if `out` is empty.
+    #[corresponds(EVP_PKEY_get_raw_public_key)]
+    pub fn raw_public_key(&self, out: &mut [u8]) -> Result<usize, ErrorStack> {
```
nox: I would rather have two methods, `raw_public_key` and `raw_public_key_len`, rather than piggyback `raw_public_key` to do a different thing when `out` is empty.

nox (changes requested):

boring/src/pkey.rs
nox: With 2 separate methods, you can change `raw_{public,private}_key` to the following signature:

```rust
fn <'a>(&self, out: &'a mut [u8]) -> Result<&'a [u8], ErrorStack>
```

kornelski:
This is a useful functionality, but the API is *very* C-like. Getting the length with `&mut []` looks odd:

```rust
buffer.resize(pk.raw_public_key(&mut [])?, 0);
let len = pk.raw_public_key(&mut buffer)?;
let key = &buffer[..len];
```

Since we don't already have a great consistency, I suggest improving the approach further.

If you add `raw_public_key_len()`, and make the `raw_public_key()` require a non-empty buffer and return the slice trimmed to the correct length:

```rust
fn raw_public_key<'buf>(&self, buf: &'buf mut [u8]) -> Result<&'buf [u8], ErrorStack> {
    …
    Ok(&buf[..size])
}
```

then the API becomes nicer to use:

```rust
let key = pk.raw_public_key(&mut buffer)?; // assuming we know the size
```

and:

```rust
buffer.resize(pk.raw_public_key_len()?, 0);
let key = pk.raw_public_key(&mut buffer)?;
```


I've explored different variations like `Result<Either<&[u8], usize>, ErrorStack>` or `Result<&[u8], Result<usize, ErrorStack>>`, or taking a closure that configures a buffer when needed, but having a separate `raw_public_key_len` method ends up being the simplest.

boring/src/pkey.rs
```diff
@@ -222,6 +222,38 @@ where
     {
         unsafe { ffi::EVP_PKEY_cmp(self.as_ptr(), other.as_ptr()) == 1 }
     }
+
+    /// Returns the length of the "raw" form of the public key. Only supported for certain key types.
+    ///
+    /// Returns the used portion of `out`.
```
kornelski:
```suggestion
```

boring/src/pkey.rs
```diff
@@ -260,6 +292,38 @@ where
         private_key_to_der_pkcs8_passphrase,
         ffi::i2d_PKCS8PrivateKey_bio
     }
+
+    /// Returns the length of the "raw" form of the private key. Only supported for certain key types.
+    ///
+    /// Returns the used portion of `out`.
```
kornelski:
```suggestion
```

kornelski (approved):

nox (approved):

nox:
Now I'm wondering, should the input be of type `&mut [MaybeUninit<u8>]`? Or do we not care about that? Any opinion @kornelski?

kornelski:
Keys are small. `MaybeUninit` would be an overkill.

Issue #362, opened 2025-06-02
Revisit Clone impl for X509Store


nox:
Implementing Clone for X509Store has been attempted twice in #108 and #339.

The latest revert in #353 documented this shouldn't be done because of `SslContextBuilder::set_cert_store` and `SslContextBuilder::cert_store_mut`.

```rust
let mut ctx = SslContext::builder().unwrap();
let store = X509Store::new().unwrap();

ctx.set_cert_store(store.clone());

// This is bad.
let aliased_store = ctx.cert_store_mut();
```

Maybe we should just remove `cert_store_mut` and make people use `set_cert_store` only?

nox:
Another solution would be to somehow make `cert_store_mut` return something if the store's refcount is 1, but there is no method to check the refcount of any crypto object in boringssl, you can only increment it with a `_up_ref` function or decrement it through a `_free` function.

kornelski:
`X509Store` itself is immutable. Everywhere else in the library there's only a one-way path from a mutable builder to an immutable object, like `X509StoreBuilder` &rarr; `X509Store`.

The problem here is that `set_cert_store` takes `X509Store`, and casts it back to a `X509StoreBuilder`, undoing the "freezing" of the object.

It seems to me that the correct design would be for `set_cert_store` to take `X509StoreBuilder` instead. It would then have a guarantee it's a mutable non-copyable object, and there would be no issue exposing it as the same `X509StoreBuilder` again.

This could be added in a backwards-compatible way:

```rust
fn set_cert_store_mut(&mut self, store: X509StoreBuilder)
```

kornelski:
Maybe the current API can be kept if exclusive vs shared status of `X509Store` could be saved. Then the `cert_store_mut` could check the state at run time, and panic if necessary.

Unfortunately, `X509_STORE_set_ex_data` is not in boringssl, so the state needs to be stored in some other way:

* Add an `AtomicBool` field to `X509StoreBuilder` and `X509Store`. This is least hacky.
* Abuse `X509_STORE_set_cleanup` and `X509_STORE_get_cleanup`. Set a no-op cleanup function on it to mark it as cloned. The setter is not thread-safe, so it would have to use a custom `fn clone(&mut self)` rather than the `Clone` trait.
* Maintain a global list of cloned `X509Store` objects. Multi-threading bottleneck.

nox:
[...]

PR #366, opened 2025-06-04
Make X509Store shareable between contexts

kornelski:
#362

### DIFF
--- a/boring/src/ssl/mod.rs
+++ b/boring/src/ssl/mod.rs
@@ -88,7 +88,7 @@ use crate::ssl::bio::BioMethod;
 use crate::ssl::callbacks::*;
 use crate::ssl::error::InnerError;
 use crate::stack::{Stack, StackRef, Stackable};
-use crate::x509::store::{X509Store, X509StoreBuilderRef, X509StoreRef};
+use crate::x509::store::{X509Store, X509StoreBuilder, X509StoreBuilderRef, X509StoreRef};
 use crate::x509::verify::X509VerifyParamRef;
 use crate::x509::{
     X509Name, X509Ref, X509StoreContextRef, X509VerifyError, X509VerifyResult, X509,
@@ -933,6 +933,8 @@ extern "C" fn rpk_verify_failure_callback(
 /// A builder for `SslContext`s.
 pub struct SslContextBuilder {
     ctx: SslContext,
+    /// If it's not shared, it can be exposed as mutable
+    has_shared_cert_store: bool,
     #[cfg(feature = "rpk")]
     is_rpk: bool,
 }
@@ -1006,7 +1008,11 @@ impl SslContextBuilder {
     #[cfg(feature = "rpk")]
     pub unsafe fn from_ptr(ctx: *mut ffi::SSL_CTX, is_rpk: bool) -> SslContextBuilder {
         let ctx = SslContext::from_ptr(ctx);
-        let mut builder = SslContextBuilder { ctx, is_rpk };
+        let mut builder = SslContextBuilder {
+            ctx,
+            is_rpk,
+            has_shared_cert_store: false,
+        };
 
         builder.set_ex_data(*RPK_FLAG_INDEX, is_rpk);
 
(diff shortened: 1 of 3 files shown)


boring/src/ssl/mod.rs
```diff
@@ -1702,19 +1740,32 @@ impl SslContextBuilder {
...
 
+        // It's not safe to return a shared &X509Store when the builder owns it exclusively
         unsafe { X509StoreBuilderRef::from_ptr(ffi::SSL_CTX_get_cert_store(self.as_ptr())) }
     }
 
     /// Returns a mutable reference to the context's certificate store.
+    ///
+    /// ## Panics
+    ///
+    /// * If the a shared store has been set via [`set_cert_store_ref`]
```
nox: Let's not make it panic and let's wait for boring 5 to introduce a breaking change here.

boring/src/ssl/mod.rs
```diff
@@ -1702,19 +1740,32 @@ impl SslContextBuilder {
 
     /// Returns a shared reference to the context's certificate store.
     #[corresponds(SSL_CTX_get_cert_store)]
+    #[deprecated(note = "This reference is useless. Use build() instead.")]
```
nox: Let's wait for boring 5 and just remove this method instead.

boring/src/ssl/mod.rs
```diff
@@ -1219,17 +1226,48 @@ impl SslContextBuilder {
...
 
+        self.has_shared_cert_store = false;
         unsafe {
             ffi::SSL_CTX_set_cert_store(self.as_ptr(), cert_store.into_ptr());
         }
     }
 
+    /// Replaces the context's certificate store, and allows mutating the store afterwards.
+    #[corresponds(SSL_CTX_set_cert_store)]
+    pub fn set_cert_store_mut(&mut self, cert_store: X509StoreBuilder) {
```
nox: Let's rename this to `set_cert_store_builder`.

boring/src/ssl/mod.rs
```diff
@@ -1219,17 +1226,48 @@ impl SslContextBuilder {
...
+        unsafe {
+            ffi::SSL_CTX_set_cert_store(self.as_ptr(), cert_store.into_ptr());
+        }
+    }
+
+    /// Replaces the context's certificate store, and keeps it immutable.
+    ///
+    /// This method allows sharing the `X509Store`, but calls to `cert_store_mut` will panic.
+    #[corresponds(SSL_CTX_set_cert_store)]
+    pub fn set_cert_store_ref(&mut self, cert_store: &X509Store) {
```
nox, line 1247: Let's rename this method to `set_cert_store` and make it take a `X509Store`, there is no need to make it take a ref.

nox (changes requested):
I don't like that you introduced a panic in `cert_store_mut`, as it can make currently existing code panic now when combined with `set_cert_store`. Let's wait for boring 5.

boring/src/ssl/mod.rs
kornelski: I've made it a ref, because this makes it shareable even if `X509Store` is not `Clone`. This way it can be used on Boring v4.

boring/src/ssl/mod.rs
kornelski: This will not panic in any existing code, because the behavior of `set_cert_store` hasn't changed. For it to panic, code needs to adopt the new `set_cert_store_ref` first.

kornelski:
Existing code won't panic, because I haven't changed the semantics of `set_cert_store`. 

It think it's better to deprecate `set_cert_store` first in v4, and then make a breaking behavior change in v5. The current API will force users to decide if they want immutable or mutable store by switching to one of the new methods.

boring/src/ssl/mod.rs
nox: Oh I see what you mean now.

nox:
I assumed that the PR also implemented Clone for X509Store hence my confusion, I'm ok with the changes now.

nox (approved):

PR #368, opened 2025-06-05
Sprinkle #[must_use]

kornelski:
This is mainly a result of `cargo clippy --fix -- -W clippy::must_use_candidate`. It's a bit spammy, but that's the consequence of the defaults Rust has picked.

### DIFF
--- a/boring/src/error.rs
+++ b/boring/src/error.rs
@@ -33,7 +33,7 @@ use crate::ffi;
 pub struct ErrorStack(Vec<Error>);
 
 impl ErrorStack {
-    /// Pop the contents of the OpenSSL error stack, and returns it.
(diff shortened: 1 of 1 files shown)


boring/src/error.rs
```diff
@@ -33,7 +33,8 @@ use crate::ffi;
 pub struct ErrorStack(Vec<Error>);
 
 impl ErrorStack {
-    /// Returns the contents of the OpenSSL error stack.
+    /// Pop the contents of the OpenSSL error stack, and returns it.
```
nox: Pops

kornelski:
```suggestion
    /// Pops the contents of the OpenSSL error stack, and returns it.
```

PR #374, opened 2025-07-21
Remove some comments referring to OpenSSL

hcstern:


rushilmehra:
> There's also a comment in https://github.com/cloudflare/boring/blob/master/hyper-boring/src/cache.rs#L79 referring to openssl and I'm pretty sure that's also a holdover but wanted to check.

Yeah we should confirm bssl behavior before just swapping out the name in the comment (it's possible the message itself is lying) so let's tackle that in follow up

PR #383, opened 2025-09-25
Clean-up legacy FIPS options

ghedo:
Per BoringSSL's FIPS policy, its `main` branch is the "update branch" for FedRAMP compliance's purposes.

This means that we can stop using a specific BoringSSL branch when enabling FIPS, as well as a number of hacks that allowed us to build more recent BoringSSL versions with an older pre-compiled FIPS modules.

In order to allow a smooth upgrade of internal projects, the `fips-compat` feature is reduced in scope and renamed to `legacy-compat-deprecated` so that we can incrementally upgrade internal BoringSSL forks. In practice this shouldn't really be something anyone else would need, since in order to work it requires a specific mix of BoringSSL version and backported patches.

### DIFF
--- a/boring-sys/build/main.rs
+++ b/boring-sys/build/main.rs
@@ -50,6 +50,7 @@ const CMAKE_PARAMS_APPLE: &[(&str, &[(&str, &str)])] = &[
         &[
             ("CMAKE_OSX_ARCHITECTURES", "arm64"),
             ("CMAKE_OSX_SYSROOT", "iphoneos"),
+            ("CMAKE_MACOSX_BUNDLE", "OFF"),
         ],
     ),
     (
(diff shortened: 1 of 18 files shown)


boring-sys/build/main.rs
```diff
@@ -340,55 +339,6 @@ fn get_boringssl_cmake_config(config: &Config) -> cmake::Config {
     boringssl_cmake
 }
 
-/// Verify that the toolchains match <https://csrc.nist.gov/CSRC/media/projects/cryptographic-module-validation-program/documents/security-policies/140sp3678.pdf>
```
bwesterb, line 343: We might still want to check we're using a recent-ish clang, instead of a specific version.

ghedo: The check is kind of annoying TBH, because it just matches strings so we'd have to parse them and hope CLang doesn't change the format going forward...

We can maybe figure something later on, but I also think that if you use an old enough Clang version it's not going to work anyway.

PR #390, opened 2025-09-29
Remove `SslCurve` API

cjpatton:
This API is incompatible with the latest internal FIPS build. Namely, the various group identifiers have been renamed since the previous version.

### DIFF
--- a/boring/src/ssl/mod.rs
+++ b/boring/src/ssl/mod.rs
@@ -2785,6 +2785,17 @@ impl SslRef {
         }
     }
 
+    /// Returns the curve ID (aka group ID) used for this `SslRef`.
+    #[corresponds(SSL_get_curve_id)]
+    #[must_use]
+    pub fn curve(&self) -> Option<u16> {
+        let curve_id = unsafe { ffi::SSL_get_curve_id(self.as_ptr()) };
+        if curve_id == 0 {
+            return None;
+        }
+        Some(curve_id)
+    }
+
     /// Returns an `ErrorCode` value for the most recent operation on this `SslRef`.
     #[corresponds(SSL_get_error)]
     #[must_use]
--- a/boring/src/ssl/test/mod.rs
+++ b/boring/src/ssl/test/mod.rs
@@ -952,6 +952,15 @@ fn sni_callback_swapped_ctx() {
 }
 
 #[test]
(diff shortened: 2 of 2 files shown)


boring/src/ssl/mod.rs
```diff
@@ -2949,17 +2828,6 @@ impl SslRef {
...
-    /// Returns the [`SslCurve`] used for this `SslRef`.
-    #[corresponds(SSL_get_curve_id)]
-    #[must_use]
-    pub fn curve(&self) -> Option<SslCurve> {
-        let curve_id = unsafe { ffi::SSL_get_curve_id(self.as_ptr()) };
-        if curve_id == 0 {
-            return None;
-        }
-        Some(SslCurve(curve_id.into()))
-    }
```
cjpatton, line 2913: It would be nice to have an alternative for this that is compatible with `set_curves_list()`.

ghedo: I think we should change this method to return the curve ID directly (i.e. `u16`), and maybe add a `fn curve_name(&self) -> Option<&str>` method that also calls `SSL_get_curve_name()`?

We'll need the former regardless, for internal use cases.

cjpatton: Sounds good!

cjpatton: Done.

boring/src/ssl/mod.rs
```diff
@@ -2887,29 +2785,15 @@ impl SslRef {
...
-                self.as_ptr(),
-                group_nids.as_ptr() as *const _,
-                group_nids.len(),
-            ))
-            .map(|_| ())
-        }
-    }
-
-    /// Returns the [`SslCurve`] used for this `SslRef`.
+    /// Returns the curve ID (aka group ID) used for this `SslRef`.
```
cjpatton, line 2788: If upstream BoringSSL is moving to replace "curve" with "group" in their API, then we might want do so here and for set_curves_list as well.

ghedo: It's not clear they are going to replace them since they still provide both names and the "group" APIs were added more than 2 years ago already. "group" is also not really a much better name given PQ... I think it was just for OpenSSL "compatibility".

PR #391, opened 2025-09-29
Remove "pq-experimental" and "underscore-wildcards" features

cjpatton:
When set, these features cause boring-sys to apply patches to the source before building BoringSSL. Instead, apply these patches unconditionally. This matches the semantcis of pre-compiled BoringSSL, where the library is expected to provide the features implemented by these patches.

We might want to do the same for "rpk", but currenlty our internal build of boringSSL doesn't appear to supply the following APIs:

```
ffi::SSL_CTX_set_nullchain_and_key
ffi::SSL_CTX_set_server_raw_public_key_certificate
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,7 @@ jobs:
       - name: Run clippy
         run: cargo clippy --all --all-targets
       - name: Check docs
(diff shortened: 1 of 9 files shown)


cjpatton:
Note: The CI failure appears to be due to an unrelated issue: https://github.com/cloudflare/boring/pull/392

PR #396, opened 2025-09-30
Add SslRef::curve_name()

ghedo:


boring/src/ssl/mod.rs
cjpatton, line 2791: Let's drop this from the public API if we don't need it right away.

```suggestion
    fn curve(&self) -> Option<u16> {
```

cjpatton (approved):

boring/src/ssl/mod.rs
ghedo: We'll need both for different internal users.

PR #399, opened 2025-10-01
Fix string data conversion in ErrorStack::put()

kornelski:


boring/src/error.rs
```diff
@@ -297,15 +299,31 @@ impl Error {
...
+    #[must_use]
+    fn data_cstr(&self) -> Option<Cow<'_, CStr>> {
+        let s = match &self.data {
+            Data::None => return None,
+            Data::CString(cstr) => return Some(Cow::Borrowed(cstr)),
+            Data::String(s) => s.as_str(),
+            Data::Static(s) => s,
+        };
+        CString::new(s).ok().map(Cow::Owned)
     }
```
toidiu, line 319: Can you add some simple docs for this.

Its a bit unfortunate that we now have two functions `data` and `data_cstr` to retrieve data. data_cstr is private for the moment (does it need to be public?), but would users need to call both functions to be sure there is no data?

Is there a way to combine the two functions? Can we return `enum Data`?

kornelski: This is intentionally private, only as a helper for put().

toidiu (approved):

rushilmehra (approved):

PR #400, opened 2025-10-02
pq patch: enable PQ by default like upstream

bwesterb:
The big diff is misleading. Applying each patch to the base 478b28ab12f and comparing them, we see:

```diff
git range-diff 478b28ab12f2001a03261624261fd041f5439706..adcd4022f75953605a9bf9f6a4a45c0b4fd8ed94 478b28ab12f2001a03261624261fd041f5439706..6f1b1e1f451e61cd2bda0922eecaa8387397ac5a
1:  adcd4022f ! 1:  6f1b1e1f4 Add additional post-quantum key agreements
    @@ Commit message
     
         This patch adds:
     
    -    1. Support for MLKEM768X25519 under the codepoint 0x11ec. The version
    -       of BoringSSL we patch against did not support it yet.
    +    1. Support for X25519MLKEM768 under the codepoint 0x11ec. The version
    +       of BoringSSL we patch against did not support it yet. Like recent
    +       upstream, enable by default.
     
         2. Supports for P256Kyber768Draft00 under 0xfe32, which we temporarily
            need for compliance reasons.  (Note that this is not the codepoint
    @@ ssl/extensions.cc: static bool tls1_check_duplicate_extensions(const CBS *cbs) {
            return true;
          default:
            return false;
    +@@ ssl/extensions.cc: bool ssl_client_hello_get_extension(const SSL_CLIENT_HELLO *client_hello,
    + }
    + 
    + static const uint16_t kDefaultGroups[] = {
    ++    SSL_GROUP_X25519_MLKEM768,
    +     SSL_GROUP_X25519,
    +     SSL_GROUP_SECP256R1,
    +     SSL_GROUP_SECP384R1,
     
      ## ssl/ssl_key_share.cc ##
     @@
```

### DIFF
--- a/boring-sys/patches/boring-pq.patch
+++ b/boring-sys/patches/boring-pq.patch
@@ -1,6 +1,6 @@
-From b98d803dbecc9d6848d8cbffa62b5c943fb75f70 Mon Sep 17 00:00:00 2001
+From 6f1b1e1f451e61cd2bda0922eecaa8387397ac5a Mon Sep 17 00:00:00 2001
 From: Bas Westerbaan <bas@cloudflare.com>
-Date: Fri, 22 Jul 2022 16:43:48 +0200
+Date: Thu, 2 Oct 2025 13:07:05 +0200
 Subject: [PATCH] Add additional post-quantum key agreements
 
 BoringSSL upstream has supported the temporary post-quantum
@@ -13,8 +13,9 @@ and many browsers are expected to switch to it before the end of 2024.
 
 This patch adds:
 
-1. Support for MLKEM768X25519 under the codepoint 0x11ec. The version
-   of BoringSSL we patch against did not support it yet.
+1. Support for X25519MLKEM768 under the codepoint 0x11ec. The version
+   of BoringSSL we patch against did not support it yet. Like recent
+   upstream, enable by default.
 
 2. Supports for P256Kyber768Draft00 under 0xfe32, which we temporarily
    need for compliance reasons.  (Note that this is not the codepoint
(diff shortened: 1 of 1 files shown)


cjpatton:
> The big diff is misleading. 

I confirm that the diff is indeed small. In fact, all it does is add X25519MLKEM768 as the most preferred group in the defaults.

Queestion: Should we also add P256Kyber768Draft00 to the default groups, with low preference? FIPS users will override the defaults anyway, but if a FIPS client wants to talk to a non-FIPS server, we can save them an HRR by supporting P256Kyber on the server side.

bwesterb:
> Queestion: Should we also add P256Kyber768Draft00 to the default groups, with low preference?

Done that.

boring-sys/patches/boring-pq.patch
```diff
@@ -3880,11 +3880,12 @@ index b13400097..894396414 100644
        return true;
      default:
        return false;
-@@ -307,6 +311,7 @@ bool ssl_client_hello_get_extension(const SSL_CLIENT_HELLO *client_hello,
+@@ -307,6 +311,8 @@ bool ssl_client_hello_get_extension(const SSL_CLIENT_HELLO *client_hello,
  }
  
  static const uint16_t kDefaultGroups[] = {
 +    SSL_GROUP_X25519_MLKEM768,
++    SSL_GROUP_P256_KYBER768_DRAFT00,
```
cjpatton, line 3888: I would lower the preference for this curve. The main reason to add it to the default list is to avoid HRR in case the client sends a key share for it.

bwesterb: This configuration is used for both client and server. If we put this any lower, then a server will prefer X25519 over P256Kyber768.

cjpatton: Ah yes, good point!

PR #401, opened 2025-10-14
Add symbol prefixing feature for BoringSSL

BarbossHack:
## Goal

Add a prefix to all symbols in libcrypto and libssl to prevent conflicts with other OpenSSL or BoringSSL versions that might be linked in the same process.

## Why?

When statically linking both OpenSSL and BoringSSL in the same project, we encountered **duplicate symbols**, for example:

```rust
rust-lld: error: duplicate symbol: AES_encrypt
>>> defined at aes-x86_64.s:328 (crypto/aes/aes-x86_64.s:328)
>>>            libcrypto-lib-aes-x86_64.o:(AES_encrypt) in archive /home/user/debug/target/debug/deps/libopenssl_sys-aa966f6339bd703d.rlib
>>> defined at aes.c:62 (/home/user/debug/target/debug/build/boring-sys-daeba9923fc21b33/out/boringssl/src/crypto/fipsmodule/aes/aes.c:62)
 >>>            bcm.c.o:(.text.AES_encrypt+0x0) in archive /home/user/debug/target/debug/deps/libboring_sys-b3f993188e8b17ea.rlib
```

## How?

The first approach was to use the BoringSSL-provided Go script `make_prefix_headers.go`. It mostly worked, but:

- It required two compilation steps: first to **build the static libraries** and extract symbols with `read_symbols.go`, then to generate `boringssl_prefix_symbols.h` with `make_prefix_headers.go`, and finally **rebuild again the static libraries** with this header. (We could have prebuilt it [like gRPC does](https://github.com/grpc/grpc/blob/master/tools/distrib/generate_boringssl_prefix_header.sh), but I preferred a fully automated workflow.)
- Some ASM/Perl-generated symbols were missed (even if present in `boringssl_prefix_symbols.h`), which I ended up fixing with the Binutils `objcopy` tool, not very straightforward.
- It required Go to run.

Therefore, I decided to use only the Binutils tools `nm` and `objcopy` to fully automate the symbol prefixing.

Steps (build.rs):

1. List static libraries to prefix symbols in (`libssl.a` and `libcrypto.a`).
2. Use `nm` to list all exported symbols in these static libraries.
3. Generate a mapping file and use `objcopy` to prefix all symbols.
4. Update the generated bindings (`bindings.rs`) with `#[link_name]` where necessary.

## Cross-compilation

Currently, this workflow is tested for Linux and Android only.

For Android, the target-specific `nm` and `objcopy` from the NDK are used.

macOS/iOS and Windows toolchains are **not tested**. I currently don’t have access to a macOS environment and have limited experience with Windows toolchains.

## Usage

I've added an opt-in feature `prefix-symbols`, as it is not tested on macOS/iOS/Windows yet.

```bash
cargo add boring --features prefix-symbols
```

### DIFF
--- a/.github/workflows/prefix.yml
+++ b/.github/workflows/prefix.yml
@@ -1,0 +1,103 @@
+name: prefix
+
+on:
+  pull_request:
+    branches:
+      - master
+  push:
+    branches:
+      - master
+
+jobs:
+  prefix-symbols:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        include:
+          - target: x86_64-unknown-linux-gnu
+            os: ubuntu-latest
+            bin: ""
+            test: true
+            custom_env: {}
+          - target: aarch64-unknown-linux-gnu
+            os: ubuntu-latest
+            bin: ""
+            test: false
+            apt_packages: crossbuild-essential-arm64 binutils-multiarch
+            custom_env:
+              CC_aarch64_unknown_linux_gnu: aarch64-linux-gnu-gcc
+              CXX_aarch64_unknown_linux_gnu: aarch64-linux-gnu-g++
+              CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER: aarch64-linux-gnu-g++
+          - target: armv7-linux-androideabi
+            os: ubuntu-latest
+            bin: ndk
+            test: false
+            custom_env: {}
+          - target: aarch64-linux-android
+            os: ubuntu-latest
+            bin: ndk
+            test: false
+            custom_env: {}
+          - target: x86_64-linux-android
+            os: ubuntu-latest
+            bin: ndk
+            test: false
+            custom_env: {}
+          - target: i686-linux-android
+            os: ubuntu-latest
+            bin: ndk
+            test: false
+            custom_env: {}
+    defaults:
+      run:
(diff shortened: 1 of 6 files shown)


rubdos:
> It required Go to run.

FYI, I've filed an issue for this upstream: https://issues.chromium.org/issues/543471701

I did some duplicate work (mostly vibe coded, not tested cross-platform at all, ...) here https://github.com/whisperfish/boring/tree/symbol-prefix; the most interesting difference is that I pull in an environment variable to customize the prefix. I was first tempted to make it a `--cfg` flag; I think that still would feel better to me.

But maybe more interestingly, you could add a patch to get rid of the Go dependency like this https://github.com/whisperfish/boring/commit/ff0462352252398d446231c90ac3bd9bac93e3c8, to work around the upstream issue.

Sytten:
@rubdos Seems like this was accepted and being rolled out

0x676e67:
cc @BarbossHack Using BoringSSL native symbol prefix may be a good option. As for depending on Go, I do not think it is a major problem. Our fork passes its tests on Linux/Android, macOS, and Windows, and it can be built and used alongside OpenSSL. https://github.com/0x676e67/btls/pull/184

BarbossHack:
Sure, in the end, this will be the easiest approach to maintain. Thanks for the feedback 👍

PR #406, opened 2025-11-14
Release backported v4 `boring` and `boring-sys`

kornelski:
These are cherry-picked backwards-compatible changes in `boring` and `boring-sys` (but not `hyper-boring`, don't release that one).

I think we should release the bugfixes and improvements we have that can work in v4, because v5 is a semver-major upgrade and may take a while to get adopted.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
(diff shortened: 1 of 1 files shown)


rushilmehra:
Lgtm but let's drop a note somewhere about not releasing backported hyper-boring

PR #407, opened 2025-11-14
Bump version to v5, allow v5 use v4-sys

kornelski:
When https://github.com/cloudflare/boring/pull/406 is released, this will be backwards-compatible with projects using `boring-sys` v4.

This will be crucial for users of `boring`, because `boring` v4 and v5 can co-exist in the dependency tree (it's merely a bloat), but `boring-sys` v4 and v5 can't (due to linking C having globals all over the place). boring v5 with boring-sys v4 will let projects upgrade their dependencies gradually.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
-version = "5.0.0-alpha.3"
+version = "5.0.0"
 repository = "https://github.com/cloudflare/boring"
 edition = "2021"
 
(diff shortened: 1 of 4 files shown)


kornelski:
@Lekensteyn

kornelski:
Irrelevant in what sense?

You haven't released v5 to crates.io yet, so I don't think anybody has experienced yet the pain I'm trying to prevent here.

nox:
Because AFAIK we haven't bumped BoringSSL in boring-sys yet.

rushilmehra:
Do you have an example of one of the new APIs?

rushilmehra:
I think it will. If you can wait until the weekend I'll put up a PR

kornelski:
Extra functions can be made to work.

v4 sys crate can contain dummy definitions of future functions to avoid compilation errors.

or new functionality in v5 that calls new functions can be behind a feature flag. It will work if the flag isn't enabled.

kornelski:
Yeah, we could make it flexible in the other direction instead.

PR #408, opened 2025-11-20
Store a &'static SslCipherRef in SslCipher

nox:
This lets us implement Send and Sync for SslCipherRef and get it for free on SslCipher.

### DIFF
--- a/boring/src/ssl/mod.rs
+++ b/boring/src/ssl/mod.rs
@@ -69,7 +69,7 @@ use std::io;
 use std::io::prelude::*;
 use std::marker::PhantomData;
 use std::mem::{self, ManuallyDrop, MaybeUninit};
-use std::ops::{Deref, DerefMut};
+use std::ops::Deref;
 use std::panic::resume_unwind;
 use std::path::Path;
 use std::ptr::{self, NonNull};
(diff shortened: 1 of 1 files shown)


kornelski:
`impl Drop` has an effect on the borrow checker, since the instances will now always live until the end of scope.

I'm surprised the type wasn't `Copy` already, given that it's effectively a `&'static CIPHER`.

boring/src/ssl/mod.rs
```diff
@@ -2432,26 +2433,26 @@ unsafe impl ForeignType for SslCipher {
...
     type Target = SslCipherRef;
 
     fn deref(&self) -> &SslCipherRef {
-        unsafe { SslCipherRef::from_ptr(self.0) }
+        self.0
     }
 }
 
 impl DerefMut for SslCipher {
     fn deref_mut(&mut self) -> &mut SslCipherRef {
```
kornelski: Sorry I haven't noticed this before, but this combined with Copy allows mutable aliasing. 
 
If there are no usecases for mutating the ciphers in place, it'd be best to delete the derefmut

nox: Do we consider that it was wrong already to implement that, and so we just remove it?

nox: Asking because there is no difference between the code in this PR and the current code on that front, you can always get two `SslCipher` values for the same cipher and use `deref_mut` on them.

kornelski: If you could get two owned sslciphers with the same pointer, then the derefmut was wrong already. `&mut` must be unique.

nox: Added a commit that removes the impl.

Issue #415, opened 2025-12-12
Adding binding to SSL_CTX_set_strict_cipher_list in SslContextBuilder


abernardeau-wallix:
Boring SSL added the SSL_CTX_set_strict_cipher_list api a few years back this with explanation:

> First, typos in cipher lists can go undetected for a long time, and
> can have surprising consequences when silently ignored.
> 
> Second, there is a tendency to use superstition in the construction of
> cipher lists, for example by "turning off" things that do not actually
> exist.  This leads to the corrosive belief that DEFAULT and ALL ought
> not to be trusted.  This belief is false.
>

This is the function here: https://github.com/google/boringssl/blob/main/include/openssl/ssl.h#L1693 . It's not present in openssl. 

I think it make sense to expose it in the rust binding.

PR #416, opened 2025-12-12
Add boring specific api set_strict_cipher_list to SslContextBuilder

abernardeau-wallix:
Boring SSL added the SSL_CTX_set_strict_cipher_list api a few years back this with explanation:

> First, typos in cipher lists can go undetected for a long time,  
> can have surprising consequences when silently ignored.
> 
> Second, there is a tendency to use superstition in the construction of
> cipher lists, for example by "turning off" things that do not actually
> exist.  This leads to the corrosive belief that DEFAULT and ALL ought
> not to be trusted.  This belief is false.
>

This is the function here: https://github.com/google/boringssl/blob/main/include/openssl/ssl.h#L1693 . It's not present in openssl. 

I think it make sense to expose it in the rust binding.

#415

### DIFF
--- a/boring/src/ssl/mod.rs
+++ b/boring/src/ssl/mod.rs
@@ -1474,6 +1474,27 @@ impl SslContextBuilder {
         }
     }
 
+    /// Sets the list of supported ciphers for protocols before TLSv1.3 but do not
+    /// tolerate anything meaningless in the cipher list.
+    ///
+    /// The `set_ciphersuites` method controls the cipher suites for TLSv1.3 in OpenSSL.
+    /// BoringSSL doesn't implement `set_ciphersuites`.
+    /// See <https://github.com/google/boringssl/blob/main/include/openssl/ssl.h#L1685>
+    ///
+    /// See [`ciphers`] for details on the format.
+    ///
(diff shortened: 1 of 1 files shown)


boring/src/ssl/mod.rs
```diff
@@ -1474,6 +1474,27 @@ impl SslContextBuilder {
...
+    ///
+    /// The `set_ciphersuites` method controls the cipher suites for TLSv1.3 in OpenSSL.
+    /// BoringSSL doesn't implement `set_ciphersuites`.
+    /// See <https://github.com/google/boringssl/blob/main/include/openssl/ssl.h#L1685>
+    ///
+    /// See [`ciphers`] for details on the format.
+    ///
+    /// [`ciphers`]: <https://docs.openssl.org/master/man1/openssl-ciphers/>.
+    pub fn set_strict_cipher_list(&mut self, cipher_list: &str) -> Result<(), ErrorStack> {
+        let cipher_list = CString::new(cipher_list).unwrap();
```
kornelski, line 1488:
```suggestion
        let cipher_list = CString::new(cipher_list).map_err(ErrorStack::internal_error)?;
```

kornelski: I see the original also has unwrap. I'll just fix them both later.

PR #419, opened 2025-12-20
Update BoringSSL to a newer version with updated patches

nox:
The BoringSSL commit this is based on is now 91a66a59b6c1435120ff83e245d7719411294386.

RPK support has drastically changed and is now more flexible, allowing clients to accept both RPKs and X.509 certificates. 

While working on it, I noticed that how we disallowed X.509-related functions on RPK contexts was not as optimal nor as safe as it could be, which led to the second commit of this PR.

I'm going away for 8 months of parental leave on Tuesday so I guess I'll not be the person pushing the big green button on this 😁


jrose-signal:
Re: CMAKE_MSVC_RUNTIME_LIBRARY: why MultiThreadedDLL over MultiThreaded? Or allowing both? It's *probably* not a problem for us, but it does seem to be a change from 4.x.

PR #426, opened 2026-01-05
Safe clone for X509Store

### LINKED ISSUE #362: Revisit Clone impl for X509Store

kornelski:
Fixes #362

### DIFF
--- a/boring/src/ssl/mod.rs
+++ b/boring/src/ssl/mod.rs
@@ -1203,16 +1203,18 @@ impl SslContextBuilder {
         }
     }
 
-    /// Use [`set_cert_store_builder`] or [`set_cert_store_ref`] instead.
+    /// Replaces the context's certificate store, and keeps it immutable.
     ///
-    /// Replaces the context's certificate store.
+    /// This method allows sharing the `X509Store`, but calls to `cert_store_mut` will panic.
+    ///
+    /// Use [`set_cert_store_builder`] to set a mutable cert store
+    /// (there's no way to have both sharing and mutability).
     #[corresponds(SSL_CTX_set_cert_store)]
-    #[deprecated(note = "Use set_cert_store_builder or set_cert_store_ref instead")]
     pub fn set_cert_store(&mut self, cert_store: X509Store) {
         #[cfg(feature = "rpk")]
         assert!(!self.is_rpk, "This API is not supported for RPK");
 
-        self.has_shared_cert_store = false;
+        self.has_shared_cert_store = true;
         unsafe {
             ffi::SSL_CTX_set_cert_store(self.as_ptr(), cert_store.into_ptr());
         }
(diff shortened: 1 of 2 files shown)


boring/src/ssl/mod.rs
```diff
@@ -1235,14 +1237,7 @@ impl SslContextBuilder {
     /// This method allows sharing the `X509Store`, but calls to `cert_store_mut` will panic.
     #[corresponds(SSL_CTX_set_cert_store)]
     pub fn set_cert_store_ref(&mut self, cert_store: &X509Store) {
-        #[cfg(feature = "rpk")]
-        assert!(!self.is_rpk, "This API is not supported for RPK");
```
cjpatton, line 1239: Is this assertion safe to delete? Perhaps the rpk feature is incompatible with this method for some reason?

boring/src/x509/store.rs
```diff
@@ -134,6 +134,23 @@ foreign_type_and_impl_send_sync! {
     pub struct X509Store;
 }
 
+impl ToOwned for X509StoreRef {
+    type Owned = X509Store;
+
+    fn to_owned(&self) -> X509Store {
+        unsafe {
+            ffi::X509_STORE_up_ref(self.as_ptr());
```
cjpatton, line 142: if the new `X509Store` goes out of scope, we need to call `X509_STORE_free()`, correct?

boring/src/x509/store.rs
```diff
@@ -160,12 +177,15 @@ impl X509StoreRef {
...
+#[should_panic = "Shared X509Store can't be mutated"]
+fn set_cert_store_pevents_mutability() {
+    use crate::ssl::*;
+
+    let mut ctx = SslContext::builder(SslMethod::tls()).unwrap();
+    let store = X509StoreBuilder::new().unwrap().build();
+
+    ctx.set_cert_store(store.clone());
+
+    // This is bad.
```
cjpatton, line 189: Bad in what way? What do we expect to happen?

boring/src/ssl/mod.rs
kornelski: This assertion hasn't been removed. It now calls another method which has the assertion.

boring/src/x509/store.rs
kornelski: This implementation copies pattern used in many other places.

`X509_STORE_free` is handled by `Drop` of `X509Store` automatically. Note that `to_owned` is on `X509StoreRef`.

boring/src/x509/store.rs
kornelski: This is copied verbatim from #362

cjpatton: Fair enough, but it would be nice to provide a more useful comment.

PR #429, opened 2026-01-05
Keep blocking set_curves_list if kx-* is set

kornelski:
This is semantically backwards-compatible, while still avoiding compilation breakage due to Cargo features.


kornelski:
@cjpatton I need your decision on this. Merge it or close it.

cjpatton:
My vote is to revert the commit this was designed to address, and wait for v5 to apply it. Short of that, 4.20 is already cut, so I think the damage is done. This change would overcomplicate things, so I would just close it.

PR #471, opened 2026-02-10
Generate dummy mlkem bindings if mlkem.h is missing

kornelski:
Includes #470 

If C lacks MLKEM APIs, and the `mlkem` Cargo feature isn't requiring a real MLKEM support, `boring-sys` will inject an always-failing fallback to allow Rust bindings to fail at run time instead of compile time.

It uses a glob to add the fallback, because Rust always prefers real definitions over globs, so the real definitions will be used if they're available.

### DIFF
--- a/boring-sys/src/mlkem_dummy.rs
+++ b/boring-sys/src/mlkem_dummy.rs
@@ -1,0 +1,163 @@
+#![allow(unused, deprecated)]
+
+use crate::{ERR_put_error, CBB, CBS, ERR_R_FATAL};
+use std::os::raw::c_int;
+use std::process::abort;
+
+#[derive(Copy, Clone, Default)]
+pub struct UnimplementedPlaceholder {
+    pub bytes: [u8; 0],
+}
+
+#[derive(Copy, Clone, Default)]
+#[deprecated(note = "mlkem.h was missing; this API won't work")]
+pub struct MLKEM768_private_key {
+    pub opaque: UnimplementedPlaceholder,
+}
+
+#[derive(Copy, Clone, Default)]
+#[deprecated(note = "mlkem.h was missing; this API won't work")]
+pub struct MLKEM768_public_key;
+#[derive(Copy, Clone, Default)]
+#[deprecated(note = "mlkem.h was missing; this API won't work")]
+pub struct MLKEM1024_private_key {
+    pub opaque: UnimplementedPlaceholder,
+}
+#[derive(Copy, Clone, Default)]
+#[deprecated(note = "mlkem.h was missing; this API won't work")]
+pub struct MLKEM1024_public_key;
+
+#[deprecated(note = "mlkem.h was missing; this API won't work")]
+pub const MLKEM_SEED_BYTES: u32 = 0;
+#[deprecated(note = "mlkem.h was missing; this API won't work")]
+pub const MLKEM_SHARED_SECRET_BYTES: u32 = 0;
+#[deprecated(note = "mlkem.h was missing; this API won't work")]
+pub const MLKEM768_PUBLIC_KEY_BYTES: u32 = 0;
+#[deprecated(note = "mlkem.h was missing; this API won't work")]
+pub const MLKEM1024_PUBLIC_KEY_BYTES: u32 = 0;
+#[deprecated(note = "mlkem.h was missing; this API won't work")]
+pub const MLKEM768_CIPHERTEXT_BYTES: u32 = 0;
+#[deprecated(note = "mlkem.h was missing; this API won't work")]
+pub const MLKEM1024_CIPHERTEXT_BYTES: u32 = 0;
+
+#[deprecated(note = "mlkem.h was missing; this API won't work")]
+pub unsafe extern "C-unwind" fn MLKEM1024_generate_key(
+    _out_encoded_public_key: *mut u8,
+    _optional_out_seed: *mut u8,
+    _out_private_key: *mut MLKEM1024_private_key,
+) {
+    abort();
+}
+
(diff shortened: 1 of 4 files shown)


boring-sys/src/mlkem_dummy.rs
```diff
@@ -0,0 +1,190 @@
...
+#[deprecated(note = "mlkem.h was missing; this API won't work")]
+pub const MLKEM1024_PUBLIC_KEY_BYTES: u32 = 0;
+#[deprecated(note = "mlkem.h was missing; this API won't work")]
+pub const MLKEM768_CIPHERTEXT_BYTES: u32 = 0;
+#[deprecated(note = "mlkem.h was missing; this API won't work")]
+pub const MLKEM1024_CIPHERTEXT_BYTES: u32 = 0;
+
+fn fail() {
+    unsafe {
+        ERR_put_error(
```
bwesterb: This only records an error, right?  Doesn't halt the execution.

kornelski: Yeah. I've renamed it to just `put_error`.

bwesterb: Shouldn't we panic and halt execution to prevent the thing you worried about?

kornelski: The `void` functions I'm worried about panic via `unimplemented!()`, which will halt the execution one way or another (unwind or abort). Other functions that can report errors do it in the normal way by pushing an error on the stack and returning a failure code (these functions are documented to return 1 on success and 0 otherwise). I wanted this to be most gracefully-degrading as possible.

boring-sys/src/mlkem_dummy.rs
```diff
@@ -0,0 +1,184 @@
...
+    unimplemented!();
+}
+
+#[deprecated(note = "mlkem.h was missing; this API won't work")]
+pub unsafe extern "C" fn MLKEM1024_private_key_from_seed(
+    _out_private_key: *mut MLKEM1024_private_key,
+    _seed: *const u8,
+    _seed_len: usize,
+) -> ::std::os::raw::c_int {
+    put_error();
```
bwesterb: We should panic here too because `MLKEM1024_private_key_from_seed` doesn't error if you give it the right parameters and so an error check could be omitted.

boring-sys/src/mlkem_dummy.rs
```diff
@@ -0,0 +1,184 @@
...
+}
+
+#[deprecated(note = "mlkem.h was missing; this API won't work")]
+pub unsafe extern "C" fn MLKEM1024_decap(
+    _out_shared_secret: *mut u8,
+    _ciphertext: *const u8,
+    _ciphertext_len: usize,
+    _private_key: *const MLKEM1024_private_key,
+) -> ::std::os::raw::c_int {
+    put_error();
```
bwesterb: Same.

boring-sys/src/mlkem_dummy.rs
```diff
@@ -0,0 +1,184 @@
...
+}
+
+#[deprecated(note = "mlkem.h was missing; this API won't work")]
+pub unsafe extern "C" fn MLKEM768_private_key_from_seed(
+    _out_private_key: *mut MLKEM768_private_key,
+    _seed: *const u8,
+    _seed_len: usize,
+) -> ::std::os::raw::c_int {
+    put_error();
+    0
```
bwesterb: Same.

boring-sys/src/mlkem_dummy.rs
```diff
@@ -0,0 +1,184 @@
...
+
+#[deprecated(note = "mlkem.h was missing; this API won't work")]
+pub unsafe extern "C" fn MLKEM768_decap(
+    _out_shared_secret: *mut u8,
+    _ciphertext: *const u8,
+    _ciphertext_len: usize,
+    _private_key: *const MLKEM768_private_key,
+) -> ::std::os::raw::c_int {
+    put_error();
+    0
```
bwesterb: Same.

boring-sys/src/mlkem_dummy.rs
```diff
@@ -0,0 +1,184 @@
...
+}
+
+#[deprecated(note = "mlkem.h was missing; this API won't work")]
+pub unsafe extern "C" fn MLKEM768_marshal_public_key(
+    out: *mut CBB,
+    public_key: *const MLKEM768_public_key,
+) -> ::std::os::raw::c_int {
+    put_error();
+    0
```
bwesterb: Same.

boring-sys/src/mlkem_dummy.rs
```diff
@@ -0,0 +1,184 @@
...
+}
+
+#[deprecated(note = "mlkem.h was missing; this API won't work")]
+pub unsafe extern "C" fn MLKEM1024_marshal_public_key(
+    out: *mut CBB,
+    public_key: *const MLKEM1024_public_key,
+) -> ::std::os::raw::c_int {
+    put_error();
+    0
```
bwesterb: Same.

boring-sys/src/mlkem_dummy.rs
kornelski: Sigh, another C API being one cut corner away from UB and vulnerabilities. I've slapped `abort()` on all of them.

bwesterb: 🙏

bwesterb (approved):

kornelski:
Nobody complained about mlkem causing build problems, so fortunately I can skip this whole workaround.

PR #482, opened 2026-03-28
rust-openssl sync

kornelski:
I've diffed API with rust-openssl and implemented some low-hanging fruit.


rushilmehra:
Need to configure AI reviews for this repo

PR #486, opened 2026-04-13
Add EVP_AEAD-based detached AEAD module

janrueth:
Introduce a new boring::aead module with Algorithm, AeadCtx, and Crypter wrappers over BoringSSL EVP_AEAD APIs, including detached seal_scatter/open_gather operations.

Document usage with module guidance, a basic detached-tag example, and an advanced scatter/gather example that explains extra_in layout and decryption handling.

Add validation and tests for key, nonce, and tag constraints, AES-GCM and XChaCha20-Poly1305 round trips, and scatter/gather with extra input; export the module from boring/src/lib.rs.

### DIFF
--- a/boring/src/aead.rs
+++ b/boring/src/aead.rs
@@ -1,0 +1,549 @@
+//! Authenticated encryption with detached authentication tags.
+//!
+//! This module wraps BoringSSL's `EVP_AEAD` interface and is intended for
+//! protocols that keep ciphertext and authentication tag in separate buffers.
+//!
+//! # Overview
+//!
+//! [`AeadCtx`] is the main type. Create one with an [`Algorithm`] and a key,
+//! then use it to encrypt and decrypt:
+//!
+//! - [`AeadCtxRef::seal_in_place`] / [`AeadCtxRef::open_in_place`] — encrypt
+//!   or decrypt a buffer in place with a detached tag. These cover the common
+//!   case (TLS record framing, packet formats with explicit tag fields, etc.).
+//!
+//! - [`AeadCtxRef::seal_scatter`] / [`AeadCtxRef::open_gather`] — lower-level
+//!   scatter/gather operations for protocols that split ciphertext output across
+//!   multiple buffers.
+//!
+//! # When to use [`crate::symm`] instead
+//!
+//! If you want one-shot helpers that allocate output buffers or APIs centered
+//! on `EVP_CIPHER`, prefer [`crate::symm`], including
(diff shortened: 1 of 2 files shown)


boring/src/aead.rs
```diff
@@ -0,0 +1,517 @@
...
+                associated_data.len(),
+            ))
+        }
+    }
+}
+
+/// Convenience wrapper for in-place detached AEAD operations.
+pub struct Crypter {
+    ctx: AeadCtx,
+    max_overhead: usize,
```
kornelski: Could Crypter be just convenience methods on AeadCtx?

I presume it exists to cache these fields, but maybe that isn't necessary? There's `EVP_AEAD_CTX_tag_len` if you really need it. I don't see anything for nonce except `#define EVP_AEAD_MAX_NONCE_LENGTH 24`.

either way, the implementation verifies the lengths itself:

```
  if (in_tag_len != tag_len) {
    OPENSSL_PUT_ERROR(CIPHER, CIPHER_R_BAD_DECRYPT);
    return 0;
  }
```

so the wrapper doesn't have to.

boring/src/aead.rs
```diff
@@ -0,0 +1,517 @@
...
+                in_out.as_ptr(),
+                in_out.len(),
+                extra_in_ptr,
+                extra_in_len,
+                associated_data.as_ptr(),
+                associated_data.len(),
+            ))?;
+        }
+
+        Ok(out_tag_len)
```
kornelski: it may be helpful to return `&out_tag[..out_tag_len]`. That still returns the size, but lets the caller use the returned slice without doing buffer slicing themselves (same for other length-returning methods)

janrueth: yeah, that's probably a nicer API, will adjust!

boring/src/aead.rs
janrueth: Yeah, we can do that. I'll remove the length check here then if it's checked already!

janrueth: regarding nonce-length validation. Seems boring supports variable length nonces if the algorithm supports it. So I'm removing the check as well. I've added a test using ChaCha20-Poly1305 which doesn't support it to observe that it will fail if misused

janrueth: Pushed a change to remove the Crypter and length checks

boring/src/aead.rs
janrueth: pushed a change that returns the correctly size slice

kornelski (approved):

kornelski (approved):

PR #487, opened 2026-04-13
Add TLS 1.2 PRF module and bindings

janrueth:
Expose a new boring::prf::tls1_prf helper that wraps the TLS 1.2 PRF and document it with RFC 5246 context. Add vector tests for SHA-224, SHA-256, SHA-384, and SHA-512 to verify output compatibility.

Declare CRYPTO_tls1_prf in boring-sys as a manual extern because the symbol is exported by BoringSSL but declared only in an internal header, so bindgen does not emit it. Keep this as a small explicit declaration in lib.rs as the simplest and most robust approach for a single missing symbol, instead of adding fragile bindgen plumbing for internal headers.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,7 @@ jobs:
           path: target
           key: clippy-${{ steps.rust-version.outputs.version }}-${{ hashFiles('Cargo.lock') }}
       - name: Run clippy
-        run: cargo clippy --all --all-targets --features rpk,mlkem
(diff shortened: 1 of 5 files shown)


kornelski:
Why the -sys crate modification? bindgen should have generated the extern "C" definition for you

janrueth:
The symbol is defined in an internal header that's not looked at by bindgen. Alternative is to add a patch that exposes it to a public header but that feels more brittle than simply defining the symbol the way I did

kornelski (approved):

janrueth:
Thanks for adding the feature-flag gate!

PR #495, opened 2026-04-21
Pingora's patch disabling cert version validation for extensions

kornelski:
Needed by Pingora to unfork boring.

### DIFF
--- a/boring-sys/Cargo.toml
+++ b/boring-sys/Cargo.toml
@@ -61,7 +61,8 @@ rpk = []
 # Require mlkem.h
 mlkem = []
 
-# Violate RFC 5280 5.1.2.1 and allow certs with CRL extensions without a correct version
+# Violate RFC 5280 5.1.2.1 and allow certs with CRL extensions in X.509 version 1.
+# This feature is not guaranteed to be supported in future minor releases.
 allow-crl-extensions-bad-version = []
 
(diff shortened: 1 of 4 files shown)


johnhurt (approved):

boring-sys/Cargo.toml
```diff
@@ -61,6 +61,9 @@ rpk = []
 # Require mlkem.h
 mlkem = []
 
+# Violate RFC 5280 5.1.2.1 and allow certs with CRL extensions without a correct version
```
cjpatton: We should give ourselves some room to remove this feature without a major version bump

```suggestion
# Violate RFC 5280 5.1.2.1 and allow certs with CRL extensions in X.509 version 1.
# This feature is not guaranteed to be supported in future minor releases.
```

cjpatton (approved):
This is unfortunate but seems relatively harmless. Sometimes you just have to match OpenSSL. 

Note that OpenSSL can be configured to be as strict as BoringSSL here, but is not configured this way by default. See X509_V_FLAG_X509_STRICT in https://docs.openssl.org/master/man3/X509_VERIFY_PARAM_set_flags/#description

johnhurt:
Where did we land on this one? I was operating under the assumption that it was already merged 😅

PR #497, opened 2026-04-21
MLDSA for Pingora

kornelski:


boring/src/mldsa.rs
```diff
@@ -0,0 +1,448 @@
+//! ML-DSA (FIPS 204) post-quantum digital signature.
+//!
+//! ML-DSA-65 is the recommended parameter set, offering NIST security level 3
+//! (roughly equivalent to AES-192).
```
cjpatton: Recommended by whom? ML-DSA-44 is more likely than -65 to get used in TLS.

boring/src/mldsa.rs
```diff
@@ -0,0 +1,448 @@
...
+                            inner: PrivateKeyInner::MlDsa87(Box::new(priv_key.assume_init())),
+                        },
+                    ))
+                }
+            }
+        }
+    }
+
+    /// Regenerates a private key from a seed value.
+    pub fn from_seed(algorithm: Algorithm, seed: &MlDsaSeed) -> Result<Self, ErrorStack> {
```
cjpatton: Note: RFC 9881 specifies a slightly more complicated DER encoding of MLDSA private keys that would be incompatible with the encoding here. We might need to at least partially implemented the complicated encoding, i.e., extract the seed and fail if the seed isn't present. This is what we're doing for internal use cases for Go. (happy to provide a reference.)

cjpatton (commented):
This looks pretty good. I'll note that BoringSSL has also integrated MLDSA into X.509, but this has happened after the commit we've currently checked out. I'm working on bumping BoringSSL in #498.

boring/src/mldsa.rs
kornelski: I've removed recommendation comments

boring/src/mldsa.rs
kornelski: I assume we can add that later.

boring/src/mldsa.rs
```diff
@@ -0,0 +1,445 @@
...
+        match self {
+            Self::MlDsa44 => ffi::MLDSA44_SIGNATURE_BYTES as usize,
+            Self::MlDsa65 => ffi::MLDSA65_SIGNATURE_BYTES as usize,
+            Self::MlDsa87 => ffi::MLDSA87_SIGNATURE_BYTES as usize,
+        }
+    }
+}
+
+/// An ML-DSA public key (any parameter set).
+pub struct MlDsaPublicKey {
```
jedisct1, line 66: Could this be `Clone`able?

boring/src/mldsa.rs
```diff
@@ -0,0 +1,445 @@
...
+                    Ok(sig)
+                }
+            }
+        }
+    }
+}
+
+impl MlDsaPublicKey {
+    /// Parses a public key from its serialized form.
+    pub fn from_bytes(algorithm: Algorithm, bytes: &[u8]) -> Result<Self, ErrorStack> {
```
jedisct1: Maybe there should be a `to_bytes()` counterpart?

boring/src/mldsa.rs
```diff
@@ -0,0 +1,445 @@
...
+use crate::cvt;
+use crate::error::ErrorStack;
+use crate::ffi;
+use crate::ffi::cbs_init;
+
+/// Seed size (32 bytes, shared across all ML-DSA parameter sets).
+pub const SEED_BYTES: usize = ffi::MLDSA_SEED_BYTES as usize;
+
+/// Raw bytes of a private key seed ([`SEED_BYTES`] long).
+pub type MlDsaSeed = [u8; SEED_BYTES];
```
cjpatton: For consistency with the mlkem module
```suggestion
pub type MlDsaPrivateKeySeed = [u8; PRIVATE_KEY_SEED_BYTES];
```

boring/src/mldsa.rs
```diff
@@ -0,0 +1,445 @@
...
+            match &self.inner {
+                PrivateKeyInner::MlDsa44(key) => {
+                    let mut sig = vec![0u8; ffi::MLDSA44_SIGNATURE_BYTES as usize];
+                    cvt(ffi::MLDSA44_sign(
+                        sig.as_mut_ptr(),
+                        key.as_ref(),
+                        msg.as_ptr(),
+                        msg.len(),
+                        core::ptr::null(),
+                        0,
```
cjpatton, line 237: This is an empty context. Was that intentional?
https://boringssl.googlesource.com/boringssl/+/main/include/openssl/mldsa.h#98

I think this is fine, but in the future we may need to add methods for signing and verifying with non-empty context.

boring/src/mldsa.rs
```diff
@@ -0,0 +1,445 @@
...
+    }
+}
+
+/// An ML-DSA public key (any parameter set).
+pub struct MlDsaPublicKey {
+    algorithm: Algorithm,
+    inner: PublicKeyInner,
+}
+
+enum PublicKeyInner {
```
cjpatton, line 72: Just to confirm: We don't need `Drop` for this because OpenSSL puts MLDSA44_public_key on the stack, correct?

cjpatton (approved):
Looks good, one cosmetic thing for consistency with mlkem

boring/src/mldsa.rs
```diff
@@ -0,0 +1,445 @@
...
+    inner: PrivateKeyInner,
+}
+
+enum PrivateKeyInner {
+    MlDsa44(Box<ffi::MLDSA44_private_key>),
+    MlDsa65(Box<ffi::MLDSA65_private_key>),
+    MlDsa87(Box<ffi::MLDSA87_private_key>),
+}
+
+impl MlDsaPrivateKey {
```
jedisct1, line 97: It would be very useful to have a `public_key()` function as well.

boring/src/mldsa.rs
kornelski: Google's docs say null/0 is valid, and it's a common case. We can add `sign_with_context` later if needed.

boring/src/mldsa.rs
kornelski: Not necessarily on stack (we box it), but it's a just a bunch of bytes stored by value.

boring/src/mldsa.rs
kornelski: That's from RustCrypto's implementation. Boring's version is a bit more involved. I'll leave it for later.

lukevalenta:
Is this still needed now that we've pulled in BoringSSL's ML-DSA support? https://github.com/cloudflare/boring/pull/509

badcf00d:
> Is this still needed now that we've pulled in BoringSSL's ML-DSA support? #509

Perhaps I'm missing something, but I assume without this MR you would still need to use the FFI to generate ML-DSA signatures? This seems like a pretty nice API just for generating standalone signatures

badcf00d:
As part of this PR it might be worth adding ML-DSA to the PKEY debug formatting here:

https://github.com/cloudflare/boring/blob/931385ed41df448c82d150219f69af56e3c8399f/boring/src/pkey.rs#L332-L344

I think the Id's would be `ffi::EVP_PKEY_ML_DSA_{44,65,87}`

Currently ML-DSA PKEY's are just printed as algorithm "unknown"

bwesterb:
> Is this still needed now that we've pulled in BoringSSL's ML-DSA support? #509

Useful for ML-DSA verifying in DNSSEC

bwesterb:
In a future version we'd want to add

1. public key marshalling
2. private -> public (as suggested by @jedisct1)

boring/src/mldsa.rs
```diff
@@ -0,0 +1,464 @@
...
+            .field("algorithm", &self.algorithm)
+            .field("seed", &"[redacted]")
+            .finish()
+    }
+}
+
+impl Drop for MlDsaPrivateKey {
+    fn drop(&mut self) {
+        unsafe {
+            ffi::OPENSSL_cleanse(self.seed.as_mut_ptr().cast(), self.seed.len());
```
bwesterb, line 389: Unpacked private key isn't zeroized. I don't really care, but LLMs will report it.

bwesterb (approved):

jedisct1:
> In a future version we'd want to add
> 
> 
> 
> 1. public key marshalling
> 
> 2. private -> public (as suggested by @jedisct1)

That would be really useful. This is currently a blocker to add ML-DSA support to JWT-simple.

lukevalenta (approved):

PR #498, opened 2026-04-21
boring-sys: Include mldsa.h and bump BoringSSL to latest

cjpatton:
1. Make MLDSA available in the FFI. Hold off on safe APIs for now, as this may need some additional discussion.
2. Bump BoringSSL to latest on main, which includes X.509 support for MLDSA. Update the PQ patch accordingly.


cjpatton:
Closing in favor of #497. Will follow up with a PR for upgrading boring-sys/deps/boringssl.

PR #499, opened 2026-04-23
Bump BoringSSL to latest and update patches

cjpatton:
The goal here is to pick up MLDSA suppport in X.509 in upstream.


boring/src/ssl/mod.rs
```diff
@@ -3862,16 +3862,22 @@ impl SslRef {
...
+    ///
+    /// Only valid on a client-side `Ssl`; returns an error on server-side SSLs.
+    #[corresponds(SSL_set1_accepted_peer_cert_types)]
     #[cfg(feature = "rpk")]
     pub fn set_server_certificate_types(
         &mut self,
         types: &[CertificateType],
     ) -> Result<(), ErrorStack> {
+        if self.is_server() {
+            return Err(ErrorStack::get());
```
lukevalenta: AI review review flagged this. Seems legitimate?

> ErrorStack::get() misuse — boring/src/ssl/mod.rs:3875-3877: ErrorStack::get() is called when self.is_server() without any preceding FFI call. Per its own docs, this returns either an empty stack ("unknown BoringSSL error") or stale errors from unrelated calls on the same thread. Use ErrorStack::internal_error_str("...") instead (pattern already used elsewhere in the crate).

```suggestion
            return Err(ErrorStack::internal_error_str("called set_server_certificate_types as server"));
```

boring/src/ssl/credential.rs
lukevalenta: AI review flagged this:

> set_spki_bytes (added in this PR) uses // instead of /// for what should be a doc comment.

boring/src/ssl/mod.rs
```diff
@@ -3880,15 +3886,22 @@ impl SslRef {
...
-    #[corresponds(SSL_get0_server_certificate_types)]
+    /// Returns the list of server certificate types that this client will
+    /// accept from the server, or `None` if none are configured.
+    ///
+    /// Only valid on a client-side `Ssl`; returns `None` on server-side SSLs.
+    #[corresponds(SSL_get0_accepted_peer_cert_types)]
     #[must_use]
     #[cfg(feature = "rpk")]
     pub fn server_certificate_types(&self) -> Option<&[CertificateType]> {
+        if self.is_server() {
```
lukevalenta, line 3917: AI review flagged this. I think we can drop this `if` block.

> Behavior change: server-side restriction is asymmetric — boring/src/ssl/mod.rs:3897-3899: server_certificate_types() returns None on servers, but upstream SSL_get0_accepted_peer_cert_types is symmetric (servers can read the types they accept from clients). Also inconsistent with SslContextBuilder::set_server_certificate_types (line 2061) which has no such guard. Either drop the restriction or document it as an intentional binding-level limitation.

boring/src/ssl/mod.rs
```diff
@@ -3901,13 +3914,24 @@ impl SslRef {
...
+    /// handshake.
+    ///
+    /// Only valid on a client-side `Ssl`; returns `None` on server-side SSLs
+    /// (a server knows its own selected credential type by other means).
+    #[corresponds(SSL_get_peer_cert_type)]
     #[must_use]
     #[cfg(feature = "rpk")]
-    pub fn selected_server_certificate_type(&self) -> CertificateType {
-        unsafe { CertificateType(ffi::SSL_get_server_certificate_type_selected(self.as_ptr())) }
+    pub fn selected_server_certificate_type(&self) -> Option<CertificateType> {
```
lukevalenta, line 3945: AI review flagged this. This one doesn't really make sense to me, so feel free to ignore.

>  Minor: selected_server_certificate_type early-return ambiguity — boring/src/ssl/mod.rs:3925-3935: Per upstream docs, SSL_get_peer_cert_type returns TLSEXT_cert_type_x509 when called too early in the handshake, indistinguishable from a real X.509 negotiation. The Option<CertificateType> return shape suggests None for "unavailable." Not a regression, but worth documenting or guarding.

lukevalenta (approved):
AI flagged a few minor things, but otherwise LGTM (for what that's worth as someone not too familiar with this crate).

The RPK patch got a lot smaller as it's now mostly provided upstream.

bwesterb (approved):

boring/src/ssl/mod.rs
cjpatton: Good catch! Done.

boring/src/ssl/mod.rs
cjpatton: The FFI call was previously SSL_get0_server_certificate_types, which returns None if the caller is not the server.

Fixed the inconsistency with line 2061 by adding the server check there.

boring/src/ssl/mod.rs
cjpatton: Yeah I'm gonna ignore this one :)

boring/src/ssl/credential.rs
cjpatton: Orthogonal, but certainly doesn't hurt. Done.

boring/src/ssl/mod.rs
cjpatton: Oops, can't call is_server() on SsslContextBuilder. Won't fix.

bwesterb:
Anything preventing us from pressing the button @kornelski ?

PR #507, opened 2026-05-21
Bump BoringSSL submodule to 4a3cda40b (API version 40)

johnhurt:
Update the BoringSSL submodule from `91a66a59b` (API version 37) to `4a3cda40b` (API version 40), moving the dependency forward by 653 commits.

### boring-pq.patch
Updated context lines in `crypto/obj/obj_dat.h` and `tool/client.cc` to match the new BoringSSL source.

### rpk.patch removed
Raw Public Key (RPK) support is now native to upstream BoringSSL as of this version, so the patch is no longer needed. The upstream implementation provides equivalent functionality under different API names:

| Old (patch) | Upstream |
|---|---|
| `SSL_CREDENTIAL_new_raw_public_key(void)` | `SSL_CREDENTIAL_new_raw_public_key(EVP_PKEY*)` |
| `SSL_get0_peer_pubkey` | `SSL_get0_peer_rpk` |
| `SSL_CTX_set_server_certificate_types` | `SSL_CTX_set1_accepted_peer_cert_types` |
| `SSL_set_server_certificate_types` | `SSL_set1_accepted_peer_cert_types` |
| `SSL_get_server_certificate_type_selected` | `SSL_get_peer_cert_type` |
| `TLS_CERTIFICATE_TYPE_X509` | `TLSEXT_cert_type_x509` |
| `TLS_CERTIFICATE_TYPE_RAW_PUBLIC_KEY` | `TLSEXT_cert_type_rpk` |

**Removed from Rust API (no upstream equivalent):**
- `SslCredentialBuilder::set_spki_bytes` — upstream constructor takes `EVP_PKEY*` directly
- `SslContextRef::server_certificate_types` — no getter in upstream
- `SslRef::server_certificate_types` — no getter in upstream

**Breaking change:** `SslCredential::new_raw_public_key()` now takes `&PKeyRef<Private>` instead of no arguments.

The `rpk` Cargo feature flag is preserved and continues to gate the Rust API surface.

### boring-sys/Cargo.toml
Added `*.cpp` and `*.in` to the include list for `cargo publish`, as the new BoringSSL version includes `third_party/benchmark` files that use these extensions.


cjpatton:
I believe this is largely a duplicate of https://github.com/cloudflare/boring/pull/499. Key differences:

1. I think your PR is a bit behind that one (`4a3cda40b` is 4/26; `23827399` is 4/21). I don't much care which version we take. In fact, we might want to upgrade to the very latest on main, as it will include TLS integration of MLDSA.
2. #499 preservers the rust API for RPK. I would like to remove it as you've done in this PR, but that would mean cutting a major version if we're aiming to maintain semver.

Also, @kornelski asked us to hold of on upgrading boring until we resolve more internal blockers for bumping boring 4 to 5. The main one is quiche, which we have a PR for that is still under review: https://github.com/cloudflare/quiche/pull/2446

johnhurt:
I see. Yeah, maintaining the rust api is a good call. I can confirm that mldsa works with the boringssl commit I am setting here, but only barely. I know the commits 2 weeks prior did not work.

cjpatton:
We've merged #499, so I think we can go ahead and close this.

PR #512, opened 2026-06-19
Add `relax-rsa-key-usage` feature to restore historical RSA keyUsage 

johnhurt:
BoringSSL 5.x includes an update to openssl that changed SSL_CONFIG::enforce_rsa_key_usage to default to true, making client-side RSA leaf keyUsage mismatches a fatal handshake error (KEY_USAGE_BIT_INCORRECT). Many real-world origins serve certificates whose keyUsage does not include the bit required by the negotiated cipher suite; OpenSSL and prior BoringSSL pins accepted them.

This adds a new `relax-rsa-key-usage` Cargo feature that applies a build-time patch restoring the historical default (false). When the feature is not enabled, the BoringSSL 5.x strict behaviour is preserved. Non-RSA keyUsage enforcement is unaffected in either case.


boring-sys/Cargo.toml
```diff
@@ -75,6 +75,15 @@ allow-crl-extensions-bad-version = []
 # `BORING_BSSL{,_FIPS}_SOURCE_PATH`.
 underscore-wildcards = []
 
+# Restores the historical BoringSSL default of not enforcing RSA keyUsage
+# during TLS handshakes. BoringSSL 5.x changed `enforce_rsa_key_usage` to
```
cjpatton: The version number refers to the `boring` crate, not BoringSSL, correct? It might be more helpful to refer to the value of BORINGSSL_API_VERSION in the BoringSSL repo itself (include/openssl/base.h).

cjpatton: Likewise for the commit message for the new patch.

cjpatton (commented):
Looks good. Can you update CI to make sure the patch is tested? Also consider adding a unit test with a cert that would validate with and without this feature.

johnhurt:
Good call. I'll take care of both.

johnhurt:
Okay, those requested changes should be handled now. I'm getting a ci error, but it seems like a flake

cjpatton:
I kicked the job and it failed the same way. Looks like an issue with CI itself, probably just needs updating.

boring-sys/Cargo.toml
```diff
@@ -75,6 +75,16 @@ allow-crl-extensions-bad-version = []
...
 
+# Restores the historical BoringSSL default of not enforcing RSA keyUsage
+# during TLS handshakes. BoringSSL starting with `BORINGSSL_API_VERSION`
+# 19 (see `include/openssl/base.h`) changed `enforce_rsa_key_usage` to
+# `true`, making a client-side RSA leaf whose keyUsage does not include the
+# bit required by the negotiated cipher suite a fatal handshake error
+# (KEY_USAGE_BIT_INCORRECT). Enabling this feature applies a build-time
+# patch that sets the default back to `false`, so RSA keyUsage mismatches
+# are non-fatal. Non-RSA keyUsage enforcement is unaffected.
+relax-rsa-key-usage = []
```
cjpatton: Consider a rename that allows us to reuse this for other changes like this.
```suggestion
relax-cert-validation = []
```

boring/test/generate-rsa-key-usage-relaxed-certs.sh
cjpatton, line 1: Perhaps move this and the generated certs to a `testdata` folder

cjpatton (approved):

boring/test/generate-rsa-key-usage-relaxed-certs.sh
johnhurt: That works for me. I didn't originally because the test dir seems like it has all the test data already - https://github.com/johnhurt/boring/tree/b971e5712ed53a5ebc8fde4cd33d47709dee982a/boring/test

cjpatton: Oh you're right, ignore me then!

cjpatton:
@johnhurt I just merged https://github.com/cloudflare/boring/pull/516, which should fix CI once you rebase.

PR #513, opened 2026-06-24
Expose TLS signature algorithm and per-connection sigalg preferences

lukevalenta:
Add bindings for inspecting the signature algorithm (incl. ML-DSA) negotiated on a TLS connection, and for setting the accepted signature algorithms per-connection.

[Expose peer signature algorithm and post-quantum sig constants](https://github.com/cloudflare/boring/commit/740b2583) is net-new functionality to surface the negotiated signature algorithm. The ML-DSA constants use hardcoded IANA codepoints (rather than `ffi::SSL_SIGN_ML_DSA_*`) so the crate continues to compile against older BoringSSL versions. This has planned usage in a downstream consumer.

[Expose per-connection verify algorithm preferences](https://github.com/cloudflare/boring/commit/1755b068) adds `SslRef::set_verify_algorithm_prefs`, the per-connection counterpart to the existing `SslContextBuilder::set_verify_algorithm_prefs`. This has planned usage in a downstream consumer.

Closes #511.

### DIFF
--- a/boring/src/ssl/mod.rs
+++ b/boring/src/ssl/mod.rs
@@ -2003,6 +2003,12 @@ impl SslContextBuilder {
     }
 
     /// Sets the context's supported signature algorithms.
+    ///
+    /// Prefer [`set_verify_algorithm_prefs`](Self::set_verify_algorithm_prefs),
+    /// which takes raw IANA codepoints rather than an OpenSSL-style colon-separated
+    /// string. Note that unlike `set_sigalgs_list`, `set_verify_algorithm_prefs`
+    /// only configures the verify preference list and does not also set the
+    /// signing algorithm prefs.
     #[corresponds(SSL_CTX_set1_sigalgs_list)]
     pub fn set_sigalgs_list(&mut self, sigalgs: &str) -> Result<(), ErrorStack> {
         let sigalgs = CString::new(sigalgs).map_err(ErrorStack::internal_error)?;
@@ -2973,6 +2979,21 @@ impl SslRef {
     pub fn set_curves_list(&mut self, curves: &str) -> Result<(), ErrorStack> {
         let curves = CString::new(curves).map_err(ErrorStack::internal_error)?;
         unsafe { cvt_0i(ffi::SSL_set1_curves_list(self.as_ptr(), curves.as_ptr())).map(|_| ()) }
+    }
+
+    #[corresponds(SSL_set_verify_algorithm_prefs)]
+    pub fn set_verify_algorithm_prefs(
+        &mut self,
+        prefs: &[SslSignatureAlgorithm],
+    ) -> Result<(), ErrorStack> {
+        unsafe {
+            cvt_0i(ffi::SSL_set_verify_algorithm_prefs(
+                self.as_ptr(),
+                prefs.as_ptr().cast(),
+                prefs.len(),
+            ))
+            .map(|_| ())
+        }
     }
 
     /// Returns the curve ID (aka group ID) used for this `SslRef`.
--- a/boring/src/ssl/mod.rs
+++ b/boring/src/ssl/mod.rs
@@ -741,6 +741,43 @@ impl SslSignatureAlgorithm {
         SslSignatureAlgorithm(ffi::SSL_SIGN_RSA_PSS_RSAE_SHA512 as _);
 
     pub const ED25519: SslSignatureAlgorithm = SslSignatureAlgorithm(ffi::SSL_SIGN_ED25519 as _);
+
+    // ML-DSA codepoints are hardcoded from the IANA TLS Signature Scheme
+    // registry so that this crate continues to compile against older
+    // BoringSSL versions that predate the SSL_SIGN_ML_DSA_* defines.
+    pub const ML_DSA_44: SslSignatureAlgorithm = SslSignatureAlgorithm(0x0904);
+
+    pub const ML_DSA_65: SslSignatureAlgorithm = SslSignatureAlgorithm(0x0905);
+
+    pub const ML_DSA_87: SslSignatureAlgorithm = SslSignatureAlgorithm(0x0906);
+
+    /// Returns the name of this signature algorithm, or `None` if unknown.
+    ///
+    /// For ECDSA algorithms the TLS 1.3 form is returned
+    /// (e.g. `ecdsa_secp256r1_sha256`), not the TLS 1.2 form (`ecdsa_sha256`).
+    #[corresponds(SSL_get_signature_algorithm_name)]
+    #[must_use]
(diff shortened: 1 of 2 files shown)


boring/src/ssl/mod.rs
```diff
@@ -741,6 +741,40 @@ impl SslSignatureAlgorithm {
...
+    pub const ML_DSA_44: SslSignatureAlgorithm =
+        SslSignatureAlgorithm(ffi::SSL_SIGN_ML_DSA_44 as _);
+
+    pub const ML_DSA_65: SslSignatureAlgorithm =
+        SslSignatureAlgorithm(ffi::SSL_SIGN_ML_DSA_65 as _);
+
+    pub const ML_DSA_87: SslSignatureAlgorithm =
+        SslSignatureAlgorithm(ffi::SSL_SIGN_ML_DSA_87 as _);
+
+    /// Returns the name of this signature algorithm, or `None` if unknown.
```
cjpatton, line 754: Document that the TLS 1.3 name is returned instead of the 1.2 name.

bwesterb (approved):

boring/src/ssl/mod.rs
```diff
@@ -749,6 +783,71 @@ impl From<u16> for SslSignatureAlgorithm {
...
+    pub const SECP521R1: SslGroup = SslGroup(ffi::SSL_GROUP_SECP521R1 as _);
+
+    pub const X25519: SslGroup = SslGroup(ffi::SSL_GROUP_X25519 as _);
+
+    pub const X25519_MLKEM768: SslGroup = SslGroup(ffi::SSL_GROUP_X25519_MLKEM768 as _);
+
+    pub const X25519_KYBER768_DRAFT00: SslGroup =
+        SslGroup(ffi::SSL_GROUP_X25519_KYBER768_DRAFT00 as _);
+
+    pub const MLKEM1024: SslGroup = SslGroup(ffi::SSL_GROUP_MLKEM1024 as _);
```
cjpatton: We previously exposed a similar API for `SslCurve`, but ripped it out in v5 because it incurred a ton of tech debt. We need to carefully consider whether we want to add it back in for groups. LIkewise for signatures.

cjpatton (changes requested):

boring/src/ssl/mod.rs
cjpatton: One thing that's changed for the better is that the FIPS and non-FIPS build agree on the BoringSSL feature set. This means we don't need all of those compile time feature gates that were causing us a lot of pain before.

But there's still one more thing to keep in mind: If we commit to `X25519_KYBER768_DRAFT00` in the public API, then we can't deprecate this obsolete algorithm until boring 6. I would advocate for at least removing this one: downstream users shouldn't need to specify it anyway.

However, given the history, I think we should drop `SslGroup` unless the existing API is insufficient. Is that the case?

boring/src/ssl/mod.rs
```diff
@@ -2005,6 +2107,22 @@ impl SslContextBuilder {
         }
     }
 
+    /// Sets the context's supported signature signing algorithms.
+    #[corresponds(SSL_CTX_set_signing_algorithm_prefs)]
+    pub fn set_signing_algorithm_prefs(
```
cjpatton: What's the use case for this?

lukevalenta: Mostly just API completion. Today `set_sigalgs_list` calls the OpenSSL-compatible `SSL_CTX_set1_sigalgs_list` which sets both the preferred signing algorithm and the preferred verify algorithm and has to do an extra lookup of string names to algorithm identifiers.

BoringSSL split this into two separate functions: the `set_verify_algorithm_prefs` is definitely useful, but I don't have an immediate use case for `set_signing_algorithm_prefs`, so happy to drop.

boring/src/ssl/mod.rs
lukevalenta: That makes sense. Should be fine to drop!

boring/src/ssl/mod.rs
```diff
@@ -741,6 +741,43 @@ impl SslSignatureAlgorithm {
...
     pub const ED25519: SslSignatureAlgorithm = SslSignatureAlgorithm(ffi::SSL_SIGN_ED25519 as _);
+
+    pub const ML_DSA_44: SslSignatureAlgorithm =
+        SslSignatureAlgorithm(ffi::SSL_SIGN_ML_DSA_44 as _);
+
+    pub const ML_DSA_65: SslSignatureAlgorithm =
+        SslSignatureAlgorithm(ffi::SSL_SIGN_ML_DSA_65 as _);
+
+    pub const ML_DSA_87: SslSignatureAlgorithm =
+        SslSignatureAlgorithm(ffi::SSL_SIGN_ML_DSA_87 as _);
```
cjpatton: This creates some friction for downstream users who want a newer version of `boring` but are stuck on an older version of `BoringSSL` (e.g., because they have their own patch set they would need to rebase), then it will fail to compile.

boring/src/ssl/mod.rs
```diff
@@ -2938,6 +2981,21 @@ impl SslRef {
         unsafe { cvt_0i(ffi::SSL_set1_curves_list(self.as_ptr(), curves.as_ptr())).map(|_| ()) }
     }
 
+    #[corresponds(SSL_set_verify_algorithm_prefs)]
+    pub fn set_verify_algorithm_prefs(
+        &mut self,
+        prefs: &[SslSignatureAlgorithm],
```
cjpatton, line 2987: Ideally we can use SSL_(CTX_)_set1_sigalgs_list(), in which case we can avoid adding the new constants to the SslSignatureAlgorithm API: https://github.com/cloudflare/boring/pull/513/changes#r3468592036

boring/src/ssl/mod.rs
lukevalenta: What about hard-coding the code points instead of pulling them from BoringSSL? Then, once we think consumers are all on recent-enough BoringSSL versions, we could again pull them via ffi.

boring/src/ssl/mod.rs
lukevalenta: Let's continue this discussion in https://github.com/cloudflare/boring/pull/513#discussion_r3468870025

boring/src/ssl/mod.rs
lukevalenta: I went ahead and dropped the `set_signing_algorithm_prefs` functions to minimize API surface area.

boring/src/ssl/mod.rs
cjpatton: This works!

cjpatton (approved):

PR #514, opened 2026-06-25
aead: require &mut self for seal/open and add TLS AEAD variants

janrueth:
Change all seal/open methods on AeadCtxRef from &self to &mut self.

The TLS-specific AEADs (aes_*_gcm_tls12, aes_*_gcm_tls13) are stateful and internally mutate the EVP_AEAD_CTX during seal operations (nonce counter tracking). Using &self allows LLVM to optimize under the assumption that the context is read-only, which can corrupt nonce state and cause cryptographic failures. This was the root cause of cloudflare/quiche#2383.

Using &mut self universally (for both generic and TLS AEADs) is the simplest fix and prevents this class of bug entirely.

Also adds Algorithm constructors for the four TLS GCM variants:
- aes_128_gcm_tls12 / aes_256_gcm_tls12
- aes_128_gcm_tls13 / aes_256_gcm_tls13

These enforce strictly monotonic nonces at the AEAD level, matching the TLS 1.2 and TLS 1.3 nonce construction requirements.

Tests cover:
- Round-trip seal/open for all TLS variants
- TLS nonce monotonicity enforcement (reuse and backwards rejected)
- Contrast: generic GCM accepts non-monotonic nonces

### DIFF
--- a/boring/src/aead.rs
+++ b/boring/src/aead.rs
@@ -36,13 +36,24 @@
 //! following safe nonce practices for the selected algorithm.
 //! [`Algorithm::nonce_len`] returns the required nonce size in bytes.
 //!
+//! # Mutability
+//!
+//! All seal and open methods require `&mut self`. While the generic AEADs
+//! (AES-GCM, ChaCha20-Poly1305, etc.) are stateless after initialization,
+//! the TLS-specific AEADs are **stateful** and internally mutate the context
+//! during seal operations. Using `&mut self` universally prevents LLVM from
+//! incorrectly optimizing away state changes — an issue that has caused
+//! real-world cryptographic failures (see [quiche#2383]).
+//!
+//! [quiche#2383]: https://github.com/cloudflare/quiche/pull/2383
+//!
 //! # Example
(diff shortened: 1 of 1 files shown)


kornelski:
Have you reported this to Google?

The C header uses `const EVP_AEAD_CTX *ctx` and Google's own `bssl-crypto` uses `&self`, so it looks like UB everywhere.

BTW, `&mut self` does help avoid concurrent access by construction, but I don't think it matters from LLVM optimization perspective in this case, because mutation happens behind a C pointer which doesn't inherit restrictions of references.

janrueth:
no, I haven't reported this to google.

Not sure about the optimizer, I followed the argument in https://github.com/cloudflare/quiche/pull/2383

kornelski (approved):

kornelski:
I've made it backwards-compatible, because upgrade to v5 is causing us so much pain, I don't want to throw v6 into the mix.

janrueth:
yeah I wasn't sure about that!  sweet, thanks.

PR #524, opened 2026-08-20
RTG-5132 Bump BoringSSL

bwesterb:
Most of our patches required work:

1. pq: (a) Upstream removed Kyber. We keep X25519Kyber768 and P256Kyber768 by implementing Kyber768 on top of ML-KEM-768.

   (b) Improve documentation of what's currently supported.

2. allow-crl-extensions-bad-version: Properly document what the patch actually does. Also improve its tests.

3. relax-cert-validation: Upstream removed enforce_rsa_key_usage; inline the behaviour instead.

I tested the Kyber key exchanges of the new BoringSSL against the old both as client and server. I didn't add a Go test for it to keep the patch small (not wanting to bring in a Kyber implementation.)

### DIFF
--- a/boring-sys/patches/boring-pq.patch
+++ b/boring-sys/patches/boring-pq.patch
@@ -16,26 +16,32 @@ This patch adds:
    non post-quantum and a post-quantum keyshare if available.  These
    functions allow one to change the behaviour to only send a single
    keyshare.
+
+4. Restores X25519Kyber768Draft00 under 0x6399, removed upstream in
+   188ce3c13.
+
+Upstream deleted its Kyber implementation, so both Kyber hybrids are now
+built on ML-KEM-768, applying Kyber's key derivation
+SHAKE-256(K || SHA3-256(c)) to the ML-KEM shared secret. See
+https://words.filippo.io/mlkem768/#bonus-track-using-a-ml-kem-implementation-as-kyber-v3
 ---
  crypto/obj/obj_dat.h                |   6 +-
  crypto/obj/obj_mac.num              |   1 +
  crypto/obj/objects.txt              |   1 +
  include/openssl/nid.h               |   3 +
- include/openssl/ssl.h               |  15 ++++
- ssl/extensions.cc                   |  26 ++++---
- ssl/internal.h                      |  12 ++-
- ssl/ssl_key_share.cc                | 111 +++++++++++++++++++++++++++-
- ssl/ssl_lib.cc                      |  16 +++-
- ssl/ssl_test.cc                     |  24 +++++-
- ssl/test/runner/basic_tests.go      |   2 +
+ include/openssl/ssl.h               |  17 ++
+ ssl/extensions.cc                   |  35 ++--
+ ssl/internal.h                      |  12 +-
+ ssl/ssl_key_share.cc                | 257 ++++++++++++++++++++++++++++
+ ssl/ssl_lib.cc                      |  16 +-
+ ssl/ssl_test.cc                     |  31 +++-
+ ssl/test/fuzzer.h                   |   6 +-
  ssl/test/runner/cbc_tests.go        |   3 +
  ssl/test/runner/common.go           |   2 +-
- ssl/test/runner/curve_tests.go      |  28 +++----
- ssl/test/runner/ech_tests.go        |  24 +++++-
- ssl/test/runner/extension_tests.go  |   3 +-
- ssl/test/runner/key_update_tests.go |   6 +-
- tool/client.cc                      |   9 +++
- 18 files changed, 249 insertions(+), 43 deletions(-)
+ ssl/test/runner/curve_tests.go      |  16 ++
+ ssl/test/runner/key_update_tests.go |   2 +
+ tool/client.cc                      |   9 +
+ 16 files changed, 393 insertions(+), 24 deletions(-)
 
 diff --git a/crypto/obj/obj_dat.h b/crypto/obj/obj_dat.h
 index feb8f2d1e..7ca28d9f5 100644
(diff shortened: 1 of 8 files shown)


Lekensteyn (approved):
LGTM. I was worried about the impact of the RSA Key Usage not relaxing TLS 1.3, but that was already the status quo before.

What wasn't immediately obvious to me, and explains why the reported kind of breakage is not more widespread: X.509 certificates with RSA keys pass the key usage check either (1) when the requested key usage bit is set, or (2) when the Key Usage extension is missing.

The last case should cover the majority of RSA keys, like the one generated with `openssl req -new -x509 -keyout server.pem -out server.pem -sha256 -nodes -subj /`

boring-sys/patches/boring-pq.patch
```diff
@@ -141,21 +157,30 @@ index 055b3e025..09d6508f4 100644
 +OPENSSL_EXPORT void SSL_CTX_use_second_keyshare(SSL_CTX *ctx, int enabled);
```
Lekensteyn, line 189: Unit tests (part of `ninja run_tests`) fail unless this is applied:

```diff
--- a/include/openssl/prefix_symbols.h
+++ b/include/openssl/prefix_symbols.h
@@ -2102,2 +2102,3 @@
 #pragma redefine_extname SSL_CTX_use_psk_identity_hint BORINGSSL_ADD_USER_LABEL_AND_PREFIX(SSL_CTX_use_psk_identity_hint)
+#pragma redefine_extname SSL_CTX_use_second_keyshare BORINGSSL_ADD_USER_LABEL_AND_PREFIX(SSL_CTX_use_second_keyshare)
 #pragma redefine_extname SSL_ECH_KEYS_add BORINGSSL_ADD_USER_LABEL_AND_PREFIX(SSL_ECH_KEYS_add)
@@ -2451,2 +2452,3 @@
 #pragma redefine_extname SSL_use_psk_identity_hint BORINGSSL_ADD_USER_LABEL_AND_PREFIX(SSL_use_psk_identity_hint)
+#pragma redefine_extname SSL_use_second_keyshare BORINGSSL_ADD_USER_LABEL_AND_PREFIX(SSL_use_second_keyshare)
 #pragma redefine_extname SSL_used_hello_retry_request BORINGSSL_ADD_USER_LABEL_AND_PREFIX(SSL_used_hello_retry_request)
@@ -5239,2 +5241,3 @@
 #define SSL_CTX_use_psk_identity_hint BORINGSSL_ADD_PREFIX(SSL_CTX_use_psk_identity_hint)
+#define SSL_CTX_use_second_keyshare BORINGSSL_ADD_PREFIX(SSL_CTX_use_second_keyshare)
 #define SSL_ECH_KEYS_add BORINGSSL_ADD_PREFIX(SSL_ECH_KEYS_add)
@@ -5588,2 +5591,3 @@
 #define SSL_use_psk_identity_hint BORINGSSL_ADD_PREFIX(SSL_use_psk_identity_hint)
+#define SSL_use_second_keyshare BORINGSSL_ADD_PREFIX(SSL_use_second_keyshare)
 #define SSL_used_hello_retry_request BORINGSSL_ADD_PREFIX(SSL_used_hello_retry_request)
```

This possibly changed in commit https://github.com/google/boringssl/commit/d32c35e573fc9c8dbe0436805cd557065892de4c (committed 2026-01-27, part of 0.20260204.0).

boring-sys/patches/boring-pq.patch
Lekensteyn: To prevent this `From` line from changing after patch changes, I used `--zero-commit`:
```
git format-patch --zero-commit --no-numbered --no-signature
```

Lekensteyn: Let's prepend `SSL_GROUP_X25519_MLKEM768` here to ensure it is enabled when using `SSL_set_compliance_policy(ssl, ssl_compliance_policy_fips_202205)`.

Alternatively, we stop patching this and require our users to move away from this API.

Lekensteyn: This note is obsolete and needs to be removed, X25519MLKEM768 is already enabled by default since 0.20260713.0 (commit date 2026-07-09) via https://github.com/google/boringssl/commit/e748facbaa5ef6e53aa51c68328e2b2245325e72

boring-sys/patches/boring-pq.patch
```diff
@@ -525,12 +716,12 @@ index 6f49d12af..5e970b2b5 100644
...
++				"-curves", strconv.Itoa(int(CurveX25519)),
  			},
  		})
  	}
 diff --git a/ssl/test/runner/common.go b/ssl/test/runner/common.go
-index c241c418e..9ffafa248 100644
+index 53fd5ea99..e21047412 100644
 --- a/ssl/test/runner/common.go
 +++ b/ssl/test/runner/common.go
 @@ -2186,7 +2186,7 @@ type ProtocolBugs struct {
```
Lekensteyn: This diff below seems incorrect and should be removed:
```diff
--- a/ssl/test/runner/common.go
+++ b/ssl/test/runner/common.go
@@ -2186,7 +2186,7 @@ type ProtocolBugs struct {
        FailIfHelloRetryRequested bool

        // FailIfPostQuantumOffered will cause a server to reject a ClientHello if
-       // post-quantum curves are supported.
+       // post-quantum curves are not supported.
        FailIfPostQuantumOffered bool

        // ExpectKeyShares, if not nil, lists (in order) the curves that a ClientHello
```

The previous comment was correct. If this flag is set, the peer will reject a Client Hello if PQ curves are included (supported).

Lekensteyn (commented):
While trying to apply the boring-pq patch on boringssl, I noticed some more minor issues. See comments.

boring-sys/patches/boring-pq.patch
bwesterb: We have users of the API

boring/src/lib.rs
```diff
@@ -13,7 +13,8 @@
 //!
 //! ## BoringSSL version
 //!
-//! By default, the crate aims to statically link with the latest BoringSSL master branch.
+//! By default, the crate statically links against a pinned BoringSSL revision, vendored as a
+//! submodule at `boring-sys/deps/boringssl`. The current pin is upstream tag `0.20260813.0`.
```
cjpatton, line 17: This comment is likely to become stale. Perhaps add an AGENTS.md with a note about this.

PR #525, opened 2026-09-22
Add native HKDF bindings

punkeel:
Add `HkdfSuite`, configured with `MessageDigest`, for RFC 5869 extraction and expansion using BoringSSL’s native HKDF functions.

Extraction returns allocation-free `DigestBytes`. Expansion fills a caller-provided buffer, accepts PRKs of at least the digest output length, and enforces the RFC output limit. Both operations return `ErrorStack` on failure.

Tests cover RFC SHA-1/SHA-256 vectors, independent SHA-384/SHA-512 vectors, empty inputs, block boundaries, output limits, and short/long PRKs. A tested example derives an AES-256-GCM key and uses it to encrypt and decrypt a message.


boring/src/hkdf.rs
```diff
@@ -0,0 +1,71 @@
+//! HMAC-based Extract-and-Expand Key Derivation Function (HKDF) with SHA-256.
+//!
+//! This module implements [RFC 5869](https://www.rfc-editor.org/rfc/rfc5869.html)
+//! using the linked BoringSSL library. FIPS status depends on the linked
+//! BoringSSL build and its configuration, and this API does not establish FIPS
+//! validation by itself.
```
cjpatton: I don't think the FIPS module is relevant. Saying we're linking against BoringSSL is redundant. The RFC reference is useful, consider folding it into the paragraph above.

```suggestion
```

boring/src/hkdf.rs
cjpatton, line 1: I think an API that supports any `MessgaeDigest` would be useful. I'll send a pointer to an internal repo where I implemented something similar.

punkeel: thank you!

boring/src/hkdf.rs
```diff
@@ -0,0 +1,126 @@
...
+/// let aad = b"record header";
+/// let plaintext = b"secret message";
+/// let mut tag = [0; 16];
+/// let ciphertext = encrypt_aead(
+///     Cipher::aes_256_gcm(), &key, Some(&nonce), aad, plaintext, &mut tag,
+/// ).unwrap();
+/// let decrypted = decrypt_aead(
+///     Cipher::aes_256_gcm(), &key, Some(&nonce), aad, &ciphertext, &tag,
+/// ).unwrap();
+/// assert_eq!(decrypted, plaintext);
```
cjpatton: Not super helpful in my opinion.

```suggestion
```

boring/src/hkdf.rs
```diff
@@ -0,0 +1,126 @@
...
+                &mut prk.len,
+                self.digest.as_ptr(),
+                ikm.as_ptr(),
+                ikm.len(),
+                salt.as_ptr(),
+                salt.len(),
+            ))?;
+        }
+
+        assert_eq!(prk.len, self.prk_size());
```
cjpatton:
```suggestion
        debug_assert_eq!(prk.len, self.prk_size());
```

cjpatton (approved):

boring/src/hkdf.rs
punkeel: (was helpful to me to confirm I understood the arg+return types :D, happy to remove)

cjpatton: I see! Feel free to keep it if you like, it's a matter of taste at the end of the day

punkeel: removed 👍

